PR #1, opened 2018-07-23
initial commit

lieut-data:
I've rebased `master` back to an empty commit to allow for this monolithic review. The sample plugin exercises all hooks and webapp components, and should double as a useful testbed for QA to verify plugin functionality from release to release.

Lots of documentation inline, so I'll leave that for review.


webapp/src/actions.js
```diff
@@ -0,0 +1,31 @@
...
+    dispatch({
+        type: CLOSE_ROOT_MODAL,
+    });
+};
+
+export const mainMenuAction = openRootModal;
+export const channelHeaderButtonAction = openRootModal;
+
+export const getStatus = () => (dispatch) => {
+    fetch('/plugins/' + PluginId + '/').then((r) => r.json()).then((r) => {
```
lieut-data: I think I'm going to expose a selector in Redux to simplify operations around using the `PluginId`, e.g. building a path to the HTTP API, or resolving websocket events, etc.

webapp/webpack.config.js
```diff
@@ -0,0 +1,42 @@
...
+                        ],
+                    },
+                },
+            },
+        ],
+    },
+    externals: {
+        react: 'react',
+        redux: 'redux',
+        'react-redux': 'reactRedux',
```
lieut-data: This depends on a forthcoming webapp change to export not `react-redux` but `reactRedux`. The way Webpack externals work, it basically tries to evaluate the exported name (hitting `window` given the global context), but `react-redux` literally resolves to `NaN`.

crspeller:
Will take a more detailed look at this later. But my initial impression is that it seems a bit complicated. I am wondering if we point developers at this they will think that plugin development is really hard. When in reality the simplest plugin is only a few lines of code and a manifest.

lieut-data:
I can agree with it seeming overwhelming at first. There's a few competing interests here:
* a sample plugin illustrating all the possible hooks / registrations. I don't think there's a way around such a "fully-featured" plugin having a lot of code, though I've tried to keep each example to the bare minimum.
* a test plugin for QA to verify all the possible hooks / registrations. No real concern with code complexity here from the QA standpoint.
* a framework from which to start plugin development

I had thought the last point could be made simpler with a `make prune` command that rips out all the "samply" bits and leaves just the build framework intact, but I'm wondering if a better approach would be to introduce a `mattermost-plugin-blank` or some such instead. Open to suggestions on this front :)

lieut-data:
After discussion, we've agreed to split this into a `mattermost-plugin-demo` and a `mattermost-plugin-sample`, the former being this PR and the latter being the build system + just one server component (`e.g. ServeHTTP`) and just one webapp component (e.g. `channel header button`).

jwilander:
Yes, shallow clone is a common term to developers familiar with git

PR #2, opened 2018-07-30
Tweak makefile targets

lieut-data:
This doesn't yet address reported `Makefile` issues for non-darwin platforms, but does move us back to `dep` while we wait for `vgo` to stabilize.

I've opted not to check in `vendor/` for the `server/` subdirectory, and instead `dep ensure` on build, with a `.depensure` file created the first time. However, to keep the `manifest` tool easy-and-quick, I have checked in the `vendor/` files there. Modifying that binary wouldn't typically be part of the workflow.

I also fixed our `tar` strategy to include the plugin id in the path. This is really only used when the plugin is sideloaded manually, but keeps us in sync with how we publish plugins today.


lieut-data:
Once merged, I'll effect the same changes to `mattermost-plugin-demo`.

PR #3, opened 2018-07-31
more Makefile improvements

lieut-data:
A few more improvements that make the initial startup experience a bit better:
* warn upfront if go/dep/npm not available
* try to find dep in $GOPATH/bin if not installed globally
* only override variables if not externally defined, allowing manual override from commandline
* split up Makefile into a buid/setup.mk for simplicity

I started out here trying to fix Linux-specific issues, but didn't actually find anything that made this Makefile not-linux specific -- tested on Debian with the standard make installed. Can I get you to give this a shot on your machine to verify?


crspeller:
@lieut-data Works for me. The `env CGO_ENABLED=0` was only for `vgo` so I removed it.

PR #6, opened 2018-08-15
Adding curl as option to 'make deploy' for better linux support

jespino:


scottleedavis:
it doesn't work for me in deploy still.  I ended up removing the '(' ')' to get it working.

Hanzei:
Tested and works!

PR #8, opened 2018-08-24
Extract plugin version from manifest

Hanzei:
This allows for e.g. log messenger when the plugin starts with the current version.

I have no experience in js. Should similar changes be made to `plugin_id.js`?

### DIFF
--- a/build/manifest/main.go
+++ b/build/manifest/main.go
@@ -13,9 +13,13 @@ import (
 const PluginIdGoFileTemplate = `package main
 
 const PluginId = "%s"
+const PluginVersion = "%s"
 `
 
-const PluginIdJsFileTemplate = `export default '%s';
+const PluginIdJsFileTemplate = `export default {
+    PluginId: '%s',
+    PluginVersion: '%s',
+};
 `
 
 func main() {
@@ -30,8 +34,11 @@ func main() {
 
 	cmd := os.Args[1]
 	switch cmd {
-	case "plugin_id":
+	case "id":
 		dumpPluginId(manifest)
+
+	case "version":
+		dumpPluginVersion(manifest)
 
 	case "has_server":
 		if manifest.HasServer() {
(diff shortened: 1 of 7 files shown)


build/manifest/main.go
```diff
@@ -31,7 +32,10 @@ func main() {
 	cmd := os.Args[1]
 	switch cmd {
 	case "plugin_id":
```
Hanzei: I think this should be changed to "id" to be a simpler command. 4/5

build/manifest/main.go
```diff
@@ -76,17 +80,22 @@ func findManifest() (*model.Manifest, error) {
 	return &manifest, nil
 }
 
-// dumpPluginId writes the plugin id from the given manifest to standard out
-func dumpPluginId(manifest *model.Manifest) {
+// dumpPluginID writes the plugin id from the given manifest to standard out
+func dumpPluginID(manifest *model.Manifest) {
```
jwilander: Can we leave it as `dumpPluginId`? I know some style checkers complain about it but this is how we do it all over and we try to be consistent

Hanzei:
Thanks for the review! I understand that you want to keep things consistent. If we follow this thought we should use one file for both web and server e.g. `manifest.js` and `manifest.go` or two files for both web and server e.g. `plugin_version.js` and `plugin_id.js` and `plugin_id. go` and `plugin_version.go`. 

I would rather go with one file per apppart.

build/manifest/main.go
lieut-data: 0/5 for this PR, but I do not this isn't just about "some style checkers" but more about writing idiomatic Golang, e.g. https://github.com/golang/go/wiki/CodeReviewComments#initialisms

lieut-data:
@Hanzei, I think a generated `manifest.(js|go)` probably makes a lot of sense!

jwilander:
@Hanzei would you mind copying these changes to https://github.com/mattermost/mattermost-plugin-demo as well?

Hanzei:
@jwilander sure, I can do this

Hanzei:
See https://github.com/mattermost/mattermost-plugin-demo/pull/6

PR #9, opened 2018-08-29
Build binaries static

hanzei:
The binaries were build as dynamically linked ones. This could cause problems on some systems.

This PR fixes this by build all binaries as statically linked ones.

Before:
```bash
$ file server/dist/plugin-linux-amd64 
server/dist/plugin-linux-amd64: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, with debug_info, not stripped
```
After: 
```bash
$ file server/dist/plugin-linux-amd64 
server/dist/plugin-linux-amd64: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, with debug_info, not stripped
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,9 +68,9 @@ endif
 server: server/.depensure
 ifneq ($(HAS_SERVER),)
 	mkdir -p server/dist;
-	cd server && env GOOS=linux GOARCH=amd64 $(GO) build -o dist/plugin-linux-amd64;
-	cd server && env GOOS=darwin GOARCH=amd64 $(GO) build -o dist/plugin-darwin-amd64;
-	cd server && env GOOS=windows GOARCH=amd64 $(GO) build -o dist/plugin-windows-amd64.exe;
+	cd server && env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build -o dist/plugin-linux-amd64;
+	cd server && env CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 $(GO) build -o dist/plugin-darwin-amd64;
+	cd server && env CGO_ENABLED=0 GOOS=windows GOARCH=amd64 $(GO) build -o dist/plugin-windows-amd64.exe;
 endif
 
 ## Ensures NPM dependencies are installed without having to run this all the time.


lieut-data:
Thanks for the PR, @Hanzei! It looks like we don't compile the mattermost-server as purely statically linked either:
```
> file bin/mattermost
bin/mattermost: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, with debug_info, not stripped
```

So will this change actually fix anything? Presumably, the server would fail to run in the same environments in which the plugin fails to run? 

Do you have an example of where this has been a problem? It would be interesting to evaluate whether or not there's a case for doing this for the server as well. One downside to doing it for plugins is increased binary sizes, especially if there are multiple plugins installed.

Hanzei:
> So will this change actually fix anything? Presumably, the server would fail to run in the same environments in which the plugin fails to run?

Yes it will fix something. A user reported this issue on my plugin repo (https://github.com/matterpoll/matterpoll/issues/18). I also encountered this problem on my own server.

BTH I'm not sure what exactly the missing libraries are. I assume it has to do with different gcc versions.

> One downside to doing it for plugins is increased binary sizes, especially if there are multiple plugins installed.

This is what I also assumed. But it turns out the binary is actually a bit smaller:
```bash
$ env  GOOS=linux GOARCH=amd64 go build -o dist/plugin-linux-amd64
$ ls -s dist/plugin-linux-amd64 
18556 dist/plugin-linux-amd64
$ env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o dist/plugin-linux-amd64
$ ls -s dist/plugin-linux-amd64 
18472 dist/plugin-linux-amd64
```
----
Build with go1.11 on a linux system.

crspeller:
@Hanzei But if these don't run because of missing libraries, shouldn't the server fail to run too? Since it's built in the same way? 

Maybe I don't understand what the core issue is.

Hanzei:
Is there a make` command to build `mattermost`?

crspeller:
@Hanzei Ya here: https://github.com/mattermost/mattermost-server/blob/master/build/release.mk#L6

Hanzei:
I tried `CGO_ENABLED=0` on `mattermost` and it also produced a smaller binary:
- 38762136 bytes without `CGO_ENABLED=0`
- 38667098bytes with `CGO_ENABLED=0`

@crspeller like it said: 
>BTH I'm not sure what exactly the missing libraries are. I assume it has to do with different gcc versions.

But it fixed my problem and results in smaller binaries.
FYI: hugo also build with  `CGO_ENABLED=0`. See https://github.com/gohugoio/hugo/blob/master/goreleaser.yml

lieut-data:
@Hanzei, do you have reproduction steps for the original issue? I think we'd like to converge on understanding this problem better before enabling this build flag. If we can see the original failure reproducing and identify the types of situations that lead to it, we'll better understand when and if we want this change on all the time.

Also, while the binary is smaller in this case, I don't know if that's going to be true all the time. I could imagine a more complex plugin (with more dependencies) ending up larger as a result. Getting to the root problem here will help tremendously.

hanzei:
Short Update: I did not forgot this PR. I will come to this back again when I have the time.

@lieut-data Sure, I will try to reproduce the original problem.

hanzei:
I was able to reproduce the error.:

1. Install `go` via `linuxbrew`
2. Build plugin binary via `env GOOS=linux GOARCH=amd64 go build -o dist/plugin-linux-amd64`
3. Copy the binary to any other system
4. Try to run it 

Expected output:
```bash
$ ./plugin-linux-amd64 
This binary is a plugin. These are not meant to be executed directly.
Please execute the program that consumes these plugins, which will
load any plugins automatically
```

Observed output:
```bash
$ ./plugin-linux-amd64 
-bash: ./plugin-linux-amd64: /home/$USERNAME/.linuxbrew/lib/ld.so: bad ELF interpreter: No such file or directory
```

Go seams to try to link to a local library inside `linuxbrew`
```bash
$ ldd ./plugin-linux-amd64 
	linux-vdso.so.1 (0x00007ffc9f7ab000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f425205b000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f4251c6a000)
	/home/$USERNAME/.linuxbrew/lib/ld.so => /lib64/ld-linux-x86-64.so.2 (0x00007f425227a000)
```

crspeller:
@hanzei Interesting. Seems like the problem is with this linuxbrew go package then? I guess if you install go regularly then it would work fine?

hanzei:
Your correct. If I compile the plugin with a "normal" installation, it works fine.

crspeller:
Do we need to do anything then?

hanzei:
Let me summarize what I have found out so far.

Some parts of the go libary like `net/http` include C code using `cgo`. Because of this the executable must either be dynamic linked against a local C library or static on compile time.

This changes fixes build for some _weird_ and in some way broken go installations. But this alone is not very convincing.

But it also builds the plugin static linked. This brings us to an old discussion in programming: "Should programs be dynamic or static linked?". I see two (but small) benefits in linking static:
- Reduces executable size: There are just a few C methods needed for the golang packages. You can see this in the reduced executable size. Unless packages with a large C codebase are added, the executable will stay smaller.
- Enhanced security and stability: If there is a (security-) bug in a C library, the admin currently has to fix it by updating the local library. With this change the bug can be fixed in the build process. Since Mattermost is released every mouth I would assume this is a more reliable way to patch bugs.

I want to point out that other go projects also use this like [Hugo](https://github.com/gohugoio/hugo/blob/master/goreleaser.yml#L7), [Goreleaser](https://github.com/goreleaser/goreleaser/blob/master/.goreleaser.yml#L6) and [Task](https://github.com/go-task/task/blob/master/.goreleaser.yml#L15).

lieut-data:
Sorry for letting this languish, @hanzei -- I really appreciate your discussion points above. 

I admit I haven't dug into this very deeply, but using some Google-fu, came across this [discussion](https://www.reddit.com/r/golang/comments/46bd5h/ama_we_are_the_go_contributors_ask_us_anything/d03rmc9/) from Ian Lance Taylor (of Google/Go):

>> I don't use cgo very often (if ever), although I can see why it's useful/necessary. Still, is there a reason CGO_ENABLED=0 isn't the default value? Or, in other words, what are the disadvantages of a truly static Go binary?

> As of Go 1.5 and 1.6, the main disadvantage of CGO_ENABLED=0 is that the os/user package will not work. Because there is so much variety in how user information is represented, that package calls out to libc to get its information. With CGO_ENABLED=0, it will always return an error "Lookup not implemented".

> Also, specifically on Darwin, the crypto/x509 package won't be able to access the root certificates on the system.

> Before Go 1.5, the net package called into libc to do DNS lookups. However, in Go 1.5 the default DNS resolver is pure Go code, and it only calls into libc in unusual situations. Except on Darwin--on Darwin it again has to call into libc in all cases.

> So if you don't care about os/user, and you aren't running on Darwin, and you don't actually use cgo yourself, then by all means set CGO_ENABLED=0 for all your builds. But for most people it's probably not the best default, at least not yet.

I find it's very difficult to get advice on `CGO_ENABLED` usage from the Go documentation, and I haven't verified if this situation has changed in newer versions of Go, but it /sounds/ like `CGO_ENABLED` might make the resulting binaries ironically less portable depending on the packages used, or at least breaking in more surprising ways.

Might you be willing to do an experiment to test the above to see if the constraints above might still apply?

That being said -- and still using the force -- I'm still inclined to leave the Go defaults alone here, letting plugin authors change the defaults as they see fit, vs. introducing a potential new point of difficulty.

crspeller:
> I'm still inclined to leave the Go defaults alone here, letting plugin authors change the defaults as they see fit, vs. introducing a potential new point of difficulty.

Going to close this unless we have a compelling reason to change the default.

hanzei:
Thank you very much @lieut-data for digging into this and licking the reddit discussion. :+1: 

I'm fine with closing this for now until new problems come up.

PR #10, opened 2018-09-20
tweak plugin id/version injection

lieut-data:
* define BUNDLE_NAME after including build/setup.mk (just for clarity, not correctness)
* separate plugin id from version with dash in bundle name to match Mattermost package
* fix various js/go idioms (some pre-existing!)
* fix error message


lieut-data:
@Hanzei pointed out that we missed a demo plugin version update. We're not releasing this plugin like we are the demo plugin, so I'm unsure if we should bump the version or not. Feedback welcome!

PR #14, opened 2018-10-04
Add travis CI

hanzei:
This PR adds Ci with travis.

Note that the npm version is not specified. If this is need, I can add this.


hanzei:
@jwilander ping

PR #15, opened 2018-10-12
 Add check-style makefile target 

hanzei:
Pretty much taken from https://github.com/mattermost/mattermost-plugin-github/blob/master/Makefile


hanzei:
@lieut-data friendly ping ;)

PR #19, opened 2018-11-07
Add make help target

hanzei:
This PR adds a `make help` target similar to the one on the Mattermost server. See https://github.com/mattermost/mattermost-server/blob/master/Makefile#L534.

The current list looks like this:
```
$ make help
all                            Checks the code style, tests, builds and bundles the plugin.
apply                          Propagates plugin manifest information into the server/ and webapp/ folders as required.
bundle                         Generates a tar bundle of the plugin for install.
check-style                    Runs govet and gofmt against all packages.
clean                          Clean removes all build artifacts.
deploy                         Installs the plugin to a (development) server.
dist                           Builds and bundles the plugin.
gofmt                          Runs gofmt against all packages.
govet                          Runs govet against all packages.
server/.depensure              Ensures the server dependencies are installed.
server                         Builds the server, if it exists, including support for multiple architectures.
test                           Runs any lints and unit tests defined for the server and webapp, if they exist.
webapp/.npminstall             Ensures NPM dependencies are installed without having to run this all the time.
webapp                         Builds the webapp, if it exists.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,16 +10,14 @@ include build/setup.mk
 
 BUNDLE_NAME ?= $(PLUGIN_ID)-$(PLUGIN_VERSION).tar.gz
 
-# all, the default target, tests, builds and bundles the plugin.
-all: check-style test dist
+all: check-style test dist ## Checks the code style, tests, builds and bundles the plugin.
 
-# apply propagates the plugin id into the server/ and webapp/ folders as required.
 .PHONY: apply
-apply:
+apply: ## Propagates plugin manifest information into the server/ and webapp/ folders as required.
 	./build/bin/manifest apply
 
 .PHONY: check-style
-check-style: server/.depensure webapp/.npminstall gofmt govet
+check-style: server/.depensure webapp/.npminstall gofmt govet ## Runs govet and gofmt against all packages.
 	@echo Checking for style guide compliance
 
 ifneq ($(HAS_WEBAPP),)
@@ -27,7 +25,7 @@ ifneq ($(HAS_WEBAPP),)
 endif
 
 .PHONY: gofmt
-gofmt:
+gofmt: ## Runs gofmt against all packages.
 ifneq ($(HAS_SERVER),)
 	@echo Running gofmt
 	@for package in $$(go list ./server/...); do \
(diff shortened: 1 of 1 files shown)


hanzei:
> Nice consistency with the server `Makefile`. One thing I didn't like on the server changes was the need to put the comment on the same line as the target: this seemed like an artificial implementation detail. Might it be feasible to do something like a grep back a few lines and extract the original comments instead?

As far as I know its not possible to `grep` for then one line. See https://stackoverflow.com/a/12652676.

> Also, `server/.depensure` and `webapp/.npminstall` aren't really meant for enduser consumption, since once the file is created, they won't run anyway. Should we omit these?

Sure, will do.

lieut-data:
Something like this converts the existing `Makefile` into the needed format for the strategy above:
```
cat Makefile | grep -v '\.PHONY' | grep -B1 -E '^[a-zA-Z_.-]+:.*' | grep -v '\-\-' | tac | awk 'NR%2{printf "%s ",$0;next;}1'
```
outputting:
```
clean: # clean removes all build artifacts
test: server/.depensure webapp/.npminstall # test runs any lints and unit tests defined for the server and webapp, if they exist
deploy: dist # variables are defined, or copying the files directly to a sibling mattermost-server directory
dist: apply \ # dist builds and bundles the plugin
bundle: # bundle generates a tar bundle of the plugin for install
webapp: webapp/.npminstall # webapp builds the webapp, if it exists
server: server/.depensure # server builds the server, if it exists, including support for multiple architectures
govet:
gofmt:
check-style: server/.depensure webapp/.npminstall gofmt govet
apply: # apply propagates the plugin id into the server/ and webapp/ folders as required.
all: check-style test dist # all, the default target, tests, builds and bundles the plugin.
```

We'd still need to add comments where needed and keep it to one line's comment. Thoughts? (And we can use `awk` instead of `tac` if there's concerns there.)

lieut-data:
(Well, not quite: still need to strip out the dependencies. But you get the idea. ;))

hanzei:
Thanks for the help @lieut-data. I updated my code. It now looks like this:
```
$ make help
all                           Checks the code style, tests, builds and bundles the plugin.
apply                         Propagates plugin manifest information into the server/ and webapp/ folders as required.
bundle                        Generates a tar bundle of the plugin for install.
check-style                   Runs govet and gofmt against all packages.
clean                         Clean removes all build artifacts.
deploy                        Installs the plugin to a (development) server.
dist                          Builds and bundles the plugin.
gofmt                         Runs gofmt against all packages.
govet                         Runs govet against all packages.
server                        Builds the server, if it exists, including support for multiple architectures.
test                          Runs any lints and unit tests defined for the server and webapp, if they exist.
webapp                        Builds the webapp, if it exists.
```

hanzei:
Just a heads up: I will push the changes to the other plugin repos once they have accumulated.

PR #21, opened 2018-11-09
configuration: ignore an empty configuration struct in setConfiguration

### LINKED ISSUE #17: Error on plugin activation
When trying out a self-compiled version of the master branch of the sample plugin, there are the following errors:
- When enabling the plugin:
```
{"level":"error","ts":1541187867.5791962,"caller":"plugin/client_rpc_generated.go:67","msg":"RPC call OnConfigurationChange to plugin failed.","plugin_id":"com.mattermost.sample-plugin","error":"unexpected EOF"}
{"level":"error","ts":1541187868.0012312,"caller":"plugin/client_rpc_generated.go:67","msg":"RPC call OnConfigurationChange to plugin failed.","plugin_id":"com.mattermost.sample-plugin","error":"connection is shut down"}
{"level":"error","ts":1541187868.238017,"caller":"plugin/client_rpc_generated.go:67","msg":"RPC call OnConfigurationChange to plugin failed.","plugin_id":"com.mattermost.sample-plugin","error":"connection is shut down"}
```

- When disabling the plugin:
```
{"level":"warn","ts":1541187955.0697181,"caller":"go-plugin/client.go:371","msg":"error closing client during Kill","plugin_id":"com.mattermost.sample-plugin","wrapped_extras":"errconnection is shut down"}`
```

Mattermost Server is a 5.4.0-rc1 Docker running on a Manjaro.
Does anyone else have that issue?

lieut-data:
If the plugin leaves the configuration struct empty, go will optimize away allocations of the zero-width struct, failing the `setConfiguration` check that prevents a user from introducing race conditions.

The demo plugin has more extensive unit tests around configuration, but it also has a non-empty struct as such. I'm inclined /not/ to test this case in the sample plugin, because this is meant as a "copy and paste" way to get started, and if you add a configuration, the test would fail.

Fixes  #17

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"reflect"
+
 	"github.com/pkg/errors"
 )
 
@@ -53,6 +55,13 @@ func (p *Plugin) setConfiguration(configuration *configuration) {
 	defer p.configurationLock.Unlock()
 
 	if configuration != nil && p.configuration == configuration {
+		// Ignore assignment if the configuration struct is empty. Go will optimize the
+		// allocation for same to point at the same memory address, breaking the check
+		// above.
+		if reflect.ValueOf(*configuration).NumField() == 0 {
+			return
+		}
+
 		panic("setConfiguration called with the existing configuration")
 	}
 


lieut-data:
@hanzei, yeah it's tricky. I mean we could throw a test into the server, but we're not shipping the code under test /with/ the server: it's here (and in a slightly different way, in the demo plugin). 

And in a real-life situation, if the plugin didn't define a configuration, it would probably make the most sense to just delete `configuration.go` altogether.

hanzei:
@lieut-data I was more thinking about a set of "common" plugins which get tested on every build on the server. This would make us more aware when we break things.

zeroimpl:
I think this should be done on mattermost-plugin-demo too. One of the first things I did was remove all fields from the struct in that file, which led it to fatal. I didn't delete the file since I planned to add new configuration fields later.

hanzei:
@zeroimpl I agree. We should fix this both plugins. @lieut-data @DSchalla what do think?

lieut-data:
@hanzei, @zeroimpl -- 100% agree. Maybe we sync up the sample and demo plugins once @DSchalla's delve changes go in?

DSchalla:
Yeah that sounds good! Since the delve changes are still in the works, personally I'd be good with adding the change already now to the sample, unless there are multiple changes already queued up - in that case we can also wait to sync them up in one sweep.

PR #22, opened 2018-11-13
Go modules

hanzei:
This PR introduces Go Modules. This allows to e.g. develop the plugin outside the `GOPATH`.

This is a proof of concept and not meant to be merged yet. Everybody who is interested in Go Modules is invited to try it out and give feedback. 

Follow up from https://community.mattermost.com/core/pl/a1naj6a61tfa8mdyyhpjfdgw8o.

### Open questions

I've opted to place `go.mod` in the root directory. I would be cleaner to have a separate one for the build tool, but this would increase the work when updating dependencies.

### Todo
- [x] Fix `make gofmt`
- [x] ~~Move `build/manifest` in `build`~~ IMO it's fine in `build/manifest`
- [X] ~~Think about using `go mod vendor`~~ I don't want this


Makefile
```diff
@@ -1,9 +1,9 @@
 GO ?= $(shell command -v go 2> /dev/null)
-DEP ?= $(shell command -v dep 2> /dev/null)
 NPM ?= $(shell command -v npm 2> /dev/null)
 HTTP ?= $(shell command -v http 2> /dev/null)
 CURL ?= $(shell command -v curl 2> /dev/null)
 MANIFEST_FILE ?= plugin.json
+export GO111MODULE=on
```
cometkim: I think this should be optional by current working directory. I know many people dislike GOPATH, but I like to use GOPATH and GOPATH-like workspace even in other languages.

hanzei: Having `GO111MODULE` set to `on` should not prevent people from placing the code in `GOPATH`, does it?

cometkim: oh.... you right, please never mind. I can't remember why I thought it would prevent to use GOPATH :sob:

hanzei: No problem :wink:

hanzei:
This is ready for review.

lieut-data:
@hanzei, haven't given this a full review yet, but can you speak to removing the `vendor` directory? I /think/ that's independent of adopting go modules, and might still be best practice -- e.g. mattermost-server will almost certainly retain `vendor` even after switching. Discussion?

hanzei:
You'r correct, you can use still use `vendor` with go modules. But what is the advantage of a `vendor` directory? The mod file pins the required dependencies. And with go modules build into the `go` command, no additional software is needed. (Which is an advantage vs. `dep`)

lieut-data:
I think these arguments apply equally well to `dep`, but we still chose to `vendor` there. Personally, I like the fact that my code will build without needing to make any network requests, or if GitHub is down, or if the source is taken down, etc. I really dislike `node_modules` for all of these reasons.

hanzei:
I'm not really see the problem. You only have to make network requests after `go.mod` has changed. But if you pulled the changes in `go.mod` the next `go` command will automatically download the new dependencies.
And updating `go.mod` only happens once a month (at least for server).

The big advantage of not using `vendor` is that you don't end up with dozen copies of files all your projects.

lieut-data:
Yeah, that's fair, at least in this context. Plus, I just realized we only vendored the `manifest` tool vs. the actual server code! Sounds good to simplify.

README.md
```diff
@@ -3,7 +3,7 @@
 This plugin serves as a starting point for writing a Mattermost plugin. Feel free to base your own plugin off this repository.
 
 ## Getting Started
-Shallow clone the repository to a directory matching your plugin name:
+Shallow clone the repository to a directory outside of `$GOPATH` matching your plugin name:
```
cometkim: Is it necessary? It seems just works well in GOPATH with go1.12, or `GO111MODULE=on` flag with go1.11+

cometkim:
> The big advantage of not using vendor is that you don't end up with dozen copies of files all your projects.

100% agree with this.

Like dep, you do not have to manage many dependencies with version diff, and it's much faster. 

> I like the fact that my code will build without needing to make any network requests, or if GitHub is down, or if the source is taken down, etc. I really dislike node_modules for all of these reasons.

I think that's a problem of SaaS, not node_modules of go module. And there are many solutions for module availability: mirroring & proxying (a.k.a artifactory) and vendor directory is one of them.

At the very least, user-centric systems like plugins don't have to suffer from inconvenience for enterprise-class availability.

Even in the case of the server, the [go module can still adapt to vendoring](https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away), so we can remove unnecessary diffs from our repository and manage them explicitly.

README.md
hanzei: If you just use the makefile, this isn't necessary. But if you can't run even a simple `go test` without manually setting `GO111MODULE=on` if the repo is in GOPATH.
But you're right, this will be no issue with g1.12.

What is the advantage of keeping this in GOPATH?

cometkim: Nothing about additional advantage. I agree to move go module, It doesn't have to be GOPATH, but also it should not be restriction about GOPATH (like "outside of `$GOPATH`") because many Gophers still have GOPATH.

Personally, I already have all my projects in the GOPATH... Haha.

hanzei: I don't have a strong opinion on moving away from `GOPATH`. Let's see what the other people think.

lieut-data: Using the force here, but I'd propose not directing where the code should live at all, and relying on `GO111MODULE` to get the semantics we want "for free". I feel like this will result in fewer surprises for plugin developers who might not be familiar with the ecosystem changes that have been happening.

hanzei: > I feel like this will result in fewer surprises for plugin developers who might not be familiar with the ecosystem changes that have been happening.

I accentually thought the opposite. If you place the code outside of `GOPATH` you will have the "old behaviour" of everything working fine. But if you place it inside `GOPATH` commands like `go test` won't work without `GO111MODULE=on`. There seams to me a lot of possibilities for surprises.

hanzei:
@levb Gently reminder about this PR :wink:

README.md
lieut-data: Ah, right -- this is only a problem if we also remove vending, since otherwise the non-module-aware tooling should work just fine.

hanzei:
@crspeller Requesting your review, since you worked on Go Modules for the Server. I would like to hear your feedback, but this isn't urgent, if you are busy.

Makefile
```diff
@@ -54,22 +54,16 @@ endif
 govet:
 ifneq ($(HAS_SERVER),)
 	@echo Running govet
-	$(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
-	$(GO) vet $$(go list ./server/...)
-	$(GO) vet -vettool=$(GOPATH)/bin/shadow $$(go list ./server/...)
+	@# Workaroung because you can't install binaries without adding them to go.mod 
+	cd /tmp	&& $(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow@latest
```
crspeller:
```suggestion
	env GO111MODULE=off $(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow@latest
```

Maybe better? Seems like this old functionality is an open issue.

hanzei: Good idea :+1:

PR #23, opened 2018-12-14
Overwrite plugin when deploying

hanzei:
Test the deployment with `curl` and it works fine. Can't test `http` because I it isn't available on my environment. Could you please test?


lieut-data:
@hanzei, wondering if we should just remove `http`, as `curl` is nearly ubiquitous and familiar. I had originally implemented with `http` because I find `curl`'s syntax cryptic, but "once it's working", the curl version is probably easier to maintain.

Thoughts?

hanzei:
I don't have much experience outside the linux world. `curl` is installed on almost every linux computer. Because of this I would be fine with removing `http`. 1/5

lieut-data:
I'm in favour of that: would you mind updating this PR to just rip it out instead? It should be transparent to practically everyone.

PR #28, opened 2019-01-16
 Allow plugins to include assets in the bundle 

hanzei:
This PR allows plugins to include assets in the bundle created with `make bundle`. This is can be e.g. used to include profile pictures for users. This is mostly a proposal. The current rules are:

- Plugins define a list of files or directories by setting. These files have to be located in `assets/`. This forces plugin devs to have a unified file structure. 2/5
- The environment variable to control this is `ASSETS` in `Makefile`. We could move this to a separate file like `variables.mk`. This should allow plugin dev to just copy past update `Makefile` without overwriting there variables.
- Assets end up in `$(PLUGIN_ID)` inside the tar bundle alongside with e.g. `plugin.json`. We could instead place them in `assets/`. 0/5

This PR also changes `tar` to not longer run in verbose mode.


hanzei:
CC @kaakaa for thoughts about this

Makefile
```diff
@@ -101,7 +106,7 @@ ifneq ($(HAS_WEBAPP),)
 	mkdir -p dist/$(PLUGIN_ID)/webapp/dist;
 	cp -r webapp/dist/* dist/$(PLUGIN_ID)/webapp/dist/;
 endif
-	cd dist && tar -cvzf $(BUNDLE_NAME) $(PLUGIN_ID)
+	cd dist && tar -czf $(BUNDLE_NAME) $(PLUGIN_ID)
```
lieut-data: 3/5, I think it's useful to be explicit, instead of surprising (when it doesn't do what you want).

Makefile
```diff
@@ -93,6 +95,9 @@ bundle:
 	rm -rf dist/
 	mkdir -p dist/$(PLUGIN_ID)
 	cp $(MANIFEST_FILE) dist/$(PLUGIN_ID)/
+ifneq ($(ASSETS),)
```
lieut-data: At first I thought this was a formalized proposal for including assets fetchable via HTTP (e.g. without any explicit plugin handling), but I believe that would require dumping things into `webapp/dist/assets` as well. This actually seems more interesting than just including static files into the build for use by the server, though I can see a case for both. (Though, sometimes it's easier to just embed the static files into the binary via something like https://github.com/rakyll/statik.)

Thoughts?

Makefile
```diff
@@ -4,6 +4,8 @@ NPM ?= $(shell command -v npm 2> /dev/null)
 HTTP ?= $(shell command -v http 2> /dev/null)
 CURL ?= $(shell command -v curl 2> /dev/null)
 MANIFEST_FILE ?= plugin.json
+# You can include assets from a assets/ directory into the bundle. This can be e.g. used to include profile pictures.
+# ASSETS = somefile
```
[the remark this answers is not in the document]
lieut-data: Yeah, feels confusing to allow `assets/something` but not `assets/somethingelse`. I'd propose having `ASSETS ?= assets` here, and have it automatically copy out of that (customizable) folder without any further subpath.  You can check if the `assets` directory exists via something like (not tested):
```
ifneq ($(wildcard $(PWD)/$(ASSETS)/.),)
    # ...
endif

File: Makefile (hunk 2 above)
Comment (hanzei) [reply Makefile]: This discussion lead to the creation of https://github.com/mattermost/mattermost-server/issues/10448.

File: Makefile (hunk 3 above)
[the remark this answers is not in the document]
Comment (hanzei) [reply Makefile]: Changes made.

**Comment by hanzei** [comment]
Based on the discussion on Community I updated this PR. Please take a second look.

**Comment by hanzei** [comment]
@jasonblais Would you please review the documentation in the README?

File: README.md [hunk 4]
```diff
@@ -45,3 +45,7 @@ In production, deploy and upload your plugin via the [System Console](https://ab
 ### How do I make a server-only or web app-only plugin?
 
 Simply delete the `server` or `webapp` folders and remove the corresponding sections from `plugin.json`. The build scripts will skip the missing portions automatically.
+
+### How do I include assets in the plugin bundle?
+
+By placing them into `assets` they automatically get bundled. You can change this directory name by editing `ASSESTDIR` in `Makefile`.
```
jasonblais: Propose

```
Place them into `assets` subfolder. You can change this directory name by editing `ASSESTDIR` in `Makefile`. 
```

Makefile
```diff
@@ -95,6 +98,9 @@ bundle:
 	rm -rf dist/
 	mkdir -p dist/$(PLUGIN_ID)
 	cp $(MANIFEST_FILE) dist/$(PLUGIN_ID)/
+ifneq ($(wildcard $(ASSESTDIR)/.),)
+	cp -r $(ASSESTDIR) dist/$(PLUGIN_ID)/
```
jasonblais: Should this be `ASSESTDIR` --> `ASSETSDIR`?

hanzei: Good catch!

README.md
hanzei: Thanks!

README.md
```diff
@@ -45,3 +45,7 @@ In production, deploy and upload your plugin via the [System Console](https://ab
 ### How do I make a server-only or web app-only plugin?
 
 Simply delete the `server` or `webapp` folders and remove the corresponding sections from `plugin.json`. The build scripts will skip the missing portions automatically.
+
+### How do I include assets in the plugin bundle?
+
+Place them into `assets` subfolder. You can change this directory name by editing `ASSESTDIR` in `Makefile`.
```
jasonblais:
```suggestion
Place them into `assets` subfolder. You can change this directory name by editing `ASSETSDIR` in `Makefile`.
```

(Sorry, my typo)

Makefile
```diff
@@ -5,6 +5,9 @@ HTTP ?= $(shell command -v http 2> /dev/null)
 CURL ?= $(shell command -v curl 2> /dev/null)
 MANIFEST_FILE ?= plugin.json
 
+# You can include assets this directory into the bundle. This can be e.g. used to include profile pictures.
+ASSETSDIR ?= assets
```
lieut-data: nit: `s/ASSETSDIR/ASSETS_DIR/`?

README.md
```diff
@@ -45,3 +45,7 @@ In production, deploy and upload your plugin via the [System Console](https://ab
 ### How do I make a server-only or web app-only plugin?
 
 Simply delete the `server` or `webapp` folders and remove the corresponding sections from `plugin.json`. The build scripts will skip the missing portions automatically.
+
+### How do I include assets in the plugin bundle?
+
+Place them into `assets` subfolder. You can change this directory name by editing `ASSETSDIR` in `Makefile`.
```
lieut-data: It may not be obvious exactly where this `assets` directory must live -- I could see someone thinking it belongs in `server` or `webapp`. Might we consider either pre-creating this folder and adding a `.gitignore` to it so that it actually gets checked out?

Separately, I feel like teaching users how to override `ASSETS_DIR` at all is overkill: no objection to actually /making/ it configurable, but the use case for renaming this folder seems low and the benefit to simplifying the description much higher. I wonder if this "space" would be more useful describing how to use the assets:

> Place them into the `assets` directory. To use an asset at runtime, build the path to your asset and open as a regular file:
```
assetsPath, err := p.API.GetAssetsPath();
if err != nil {
    return errors.Wrap(err, "failed to get assets path")
}

profileImage, err := ioutil.ReadFile(filepath.Join(assetsPath, "profile_image.png"))
if err != nil {
    return errors.Wrap(err, "failed to read profile image")
}

appErr := p.API.SetProfileImage(userId, filepath.Join(assetsPath, "profile_image.png"))
if appErr != nil {
    return errors.Wrap(err, "failed to set profile image")
}
```

hanzei: :+1: for both suggestions. I added a `.gitkeep` into `assets` instead of  `.gitignore`, but I'm only 1/5 on the file name.

PR #29, opened 2019-01-16
Add make coverage

hanzei:
This PR adds a 'make coverage` command, which creates a coverage report for the server code.

The code isn't DRY but I don't know how to fix this without adding a new `test-server` target.

Is there a similar tool for webapp?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,14 @@ ifneq ($(HAS_WEBAPP),)
 	cd webapp && $(NPM) run fix;
 endif
 
+## Creates a coverage report for the server code.
+.PHONY: coverage
+coverage: server/.depensure webapp/.npminstall
+ifneq ($(HAS_SERVER),)
+	cd server && $(GO) test -race -coverprofile=coverage.txt ./...
+	@cd server && $(GO) tool cover -html=coverage.txt
+endif
+
 ## Clean removes all build artifacts.
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ endif
 .PHONY: test
 test: server/.depensure webapp/.npminstall
 ifneq ($(HAS_SERVER),)
-	cd server && $(GO) test -race -v -coverprofile=coverage.txt ./...
+	cd server && $(GO) test -race -v ./...
 endif
 ifneq ($(HAS_WEBAPP),)
 	cd webapp && $(NPM) run fix;


Makefile
```diff
@@ -142,12 +142,20 @@ endif
...
 endif
 ifneq ($(HAS_WEBAPP),)
 	cd webapp && $(NPM) run fix;
 endif
 
+## Creates a coverage report for the server code.
+.PHONY: coverage
+coverage: server/.depensure webapp/.npminstall
+ifneq ($(HAS_SERVER),)
+	cd server && $(GO) test -race -coverprofile=coverage.txt ./...
```
lieut-data: What if we always `-coverprofile=coverage.txt` in the `test` target, and make this depend on the `test` command? Using the force, I'd say the coverage overhead likely minimal for a plugin, so it makes sense to be on all the time, and this way we eliminate the duplication.

We already have a similar pattern with `make deploy` and `make dist`, in that the former /could/ just deploy and assume a build exists, but builds anyway to ensure there is never any confusion about what is being deployed. I think it makes sense for `coverage` (or `coverage-report`) to work the same.

It would be nice if the Make target explicitly listed the path at which the coverage information could be found, e.g.:
```
echo Coverage output to /absolute/path/to/coverage.html
```

make it easy to click on the path in supporting terminals and open in a browser as such.

hanzei: > What if we always `-coverprofile=coverage.txt` in the `test` target, and make this depend on the `test` command? Using the force, I'd say the coverage overhead likely minimal for a plugin, so it makes sense to be on all the time, and this way we eliminate the duplication.

But this way we run the webapp tests every time we want to get the coverage. This can take some time and is inconvenient.
We could add a `test-server` target and run this before `coverage`.

> make it easy to click on the path in supporting terminals and open in a browser as such.

`go tool cover` already opens the browser with the file. At least on linux. Could you test, if it works on MacOS as well?

kaakaa: On macOS, coverage file is opened in browder automatically after running `make coverage`.

lieut-data: Oh, interesting re: auto-opening coverage file. Seems good enough!

0/5, but how often would one run `make coverage` -- would the time spent running the tests matter much? And presumably if/when we have webapp coverage, we'd want to generate that too anyway?

hanzei: For me it isn't worth to let plugin devs wait for a test result just to get your code DRY. @lieut-data Would you be okay, if we merge it as it is?

lieut-data: So I'm not proposing to do this just to get the code DRY, but actually that `make coverage` should always run the tests. I can't imagine a scenario where I'm iterating with `make coverage` and don't want the latest coverage data to be generated, but can imagine a developer incorrectly iterating with `make coverage` without running `make test` first.

I originally had this same idea back when `make build` and `make deploy` were separate targets, but @crspeller convinced me that it made more sense for `make deploy` to explicitly `make build` so that it always deployed the latest code. In practice, this has saved me from accidentally iterating on a plugin whilst forgetting to build it, and I'm inclined to  do the same for `make coverage` here for the same reasons.

(Don't forget `.gitignore` too to ignore `coverage.txt`.)

lieut-data: Ah, I see, I'm just getting my wires crossed. I think I'm stuck thinking about a future where `make coverage` does something useful for both webapp and server, and isn't just server specific.

I agree that the current implementation seems reasonable, and there really isn't anything preventing us from changing the semantics in the future with an appropriate change log, so let's get this merged as-is :)

PR #31, opened 2019-02-28
introducing circleci

cpanato:
Based on the dev meeting we are moving to circleci


Makefile
```diff
@@ -52,7 +52,9 @@ endif
 govet:
 ifneq ($(HAS_SERVER),)
 	@echo Running govet
-	@$(GO) vet -shadow $$(go list ./server/...) || exit 1
+	$(GO) get -u golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
+	$(GO) vet $(GOFLAGS) $$(go list ./server/...) || exit 1
+	$(GO) vet -vettool=$(which shadow) $$(go list ./server/...) || exit 1
```
hanzei: @lieut-data provided a PR to change this on the server. (https://github.com/mattermost/mattermost-server/pull/10389) Should this also be changed here?

lieut-data: Yes! I found that `$(which shadow)` doesn't actually find `shadow` at all to my dismay (at least in the context of the `Makefile` -- bash works fine). So I've hard-coded the path, and improved other things in the linked PR.

I'm also debating the `-u` in the `go get` -- it makes the process 1s slower than it feels like it ought to be. Thoughts about not forcing the upgrade each time?

hanzei: Let's continue this discussion on https://github.com/mattermost/mattermost-server/pull/10389/

README.md
```diff
@@ -1,4 +1,4 @@
-# Sample Plugin [![Build Status](https://travis-ci.org/mattermost/mattermost-plugin-sample.svg?branch=master)](https://travis-ci.org/mattermost/mattermost-plugin-sample)
+# Sample Plugin [![CircleCI](https://circleci.com/gh/mattermost/mattermost-plugin-sample.svg?style=svg)](https://circleci.com/gh/mattermost/mattermost-plugin-sample)
```
hanzei: Nit: What do you think about using https://shields.io/ instead for the unified style of the badge?

cpanato: we can use nothing against

Makefile
hanzei: @cpanato Please update to 

```suggestion
	$(GO) vet -vettool=$(GOPATH)/bin/shadow $$(go list ./server/...) || exit 1
```

as discussed in https://github.com/mattermost/mattermost-server/pull/10389

cpanato:
@hanzei PTAL

PR #33, opened 2019-03-19
MM-14575: Automatically serve static files for plugins

happygaijin:
* Added sample public file
* Added HAS_PUBLIC to Makefile to test for public folder
* Added "static_files" config setting
* Related to:
 [https://github.com/mattermost/mattermost-server/pull/10476](https://github.com/mattermost/mattermost-server/pull/10476l)
[https://github.com/mattermost/mattermost-server/issues/10447](https://github.com/mattermost/mattermost-server/issues/10447)


Makefile
```diff
@@ -75,6 +75,10 @@ ifneq ($(HAS_SERVER),)
 	cd server && env GOOS=linux GOARCH=amd64 $(GO) build -o dist/plugin-linux-amd64;
 	cd server && env GOOS=darwin GOARCH=amd64 $(GO) build -o dist/plugin-darwin-amd64;
 	cd server && env GOOS=windows GOARCH=amd64 $(GO) build -o dist/plugin-windows-amd64.exe;
+ifneq ($(HAS_PUBLIC),)
+	mkdir -p server/dist/public;
```
hanzei: What do you think about pre creating this, similar to `assets` https://github.com/mattermost/mattermost-plugin-sample/tree/master/assets?

Makefile
```diff
@@ -75,6 +75,10 @@ ifneq ($(HAS_SERVER),)
...
+	cp -r server/public/* server/dist/public/;
```
hanzei: This should be part of `bundle`

build/setup.mk
```diff
@@ -25,6 +25,9 @@ HAS_SERVER ?= $(shell build/bin/manifest has_server)
 # Determine if a webapp is defined in the manifest.
 HAS_WEBAPP ?= $(shell build/bin/manifest has_webapp)
 
+# Determine if a /public folder is in use
+HAS_PUBLIC ?= $(wildcard server/public/*)
```
hanzei: Would you please change this to 
```suggestion
HAS_PUBLIC ?= $(wildcard public/.)
```

Makefile
happygaijin: Yeah, especially with getting rid of the static_files setting in the plugin.json. It makes sense.

build/setup.mk
happygaijin: Sure. The destination on the server would then just be the root of the plugin folder?

hanzei: If you are talking about the destination in the bundle: Yes.

happygaijin: 👍

levb:
@happygaijin @hanzei This can be merged regardless of the other PRs, right?

hanzei:
@levb Yes

PR #34, opened 2019-03-22
Add .editorconfig

hanzei:
Downstream from https://github.com/mattermost/mattermost-plugin-nps/pull/5

Credits to @hmhealey for creating this file!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,28 @@
+# http://editorconfig.org/
+
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+trim_trailing_whitespace = true
+
+[*.go]
+indent_style = tab
+
+[*.{js,jsx,json,html}]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,28 @@
...
+[webapp/package.json]
+indent_size = 2
+
+[Makefile,*.mk]
+indent_style = tab
+
+[*.md]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = false
```
lieut-data: Is there a specific reason this is `false`?

hanzei: You can do linebreaks in `md` by ending a line with two spaces. Some people use this and we might want to support it.

lieut-data: Fascinating, I had no idea! Seems both useful and evil...

hanzei: I'm not 100% sure that this is a good part of markdown...

PR #39, opened 2019-05-06
[MM-15425] Improve plugin deploy/redeploy/debug cycle

cpoile:
Improve the workflow for plugin development. 

make debug command:
- only recompile for the architecture the developer is actually using (speeds up the change/test cycle)
- compile with debug flags for the server plugin (allows us to attach a dlv process and debug the server component. I'll write instructions.)

make webapp-debug command:
- compile in dev mode (with source maps) for the webapp
- setup npm correctly so that webapp changes are compiled immediately, and only changes need to be compiled
- symlink the webapp directory so that webapp changes don't need to redeployed (a plugin restart will pick up the changes immediately)
  - this means that the webapp can be debugged directly from the development directory (yay)

make reset command:
- disable the plugin
- enable the plugin (to pick up new changes in the server and webapp)
  - this stops the current instance of the plugin (to prevent 20+ zombie plugin processes)

make stop command:
- stop the webapp watcher

Quick Instructions (I'll write more detailed instructions when I merge this into the demo plugin... so that everyone can use this)
1. Instead of running `make deploy`, run `make debug`. 
    - It will set up the link between the mattermost server and the plugin's directory.
1. Run `make webapp-debug` to build the webapp with source maps, and start the npm watch process.
1. Whenever a change is made to the plugin's server component, run `make debug` to recompile and reload the plugin.
1. Whenever a change is made to the webapp, run `make reset` to stop and restart the plugin process on the server. Your changes will now appear (you may need to refresh the page)
1. To debug the server part of the plugin, run the script `./debug-plugin.sh` which will attach a delve headless process to the plugin. Attach your debugger of choice to localhost:2346 
1. To debug the webapp part of the plugin, attach the debugger to localhost:8065 (as you would normally debug the mattermost webapp).

## Links
[MM-15425](https://mattermost.atlassian.net/browse/MM-15425)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,6 +177,9 @@ debug: server-debug reset
 
 .PHONY: server-debug
 server-debug:
+ifeq ($(wildcard ../mattermost-server/.*),)
+	$(error in order to use the make debug commands, your plugin's development directory needs to be located in the same parent directory as the mattermost-server (they need to be 'peers' or 'siblings').)
+endif
 
 	./build/bin/manifest apply
 	mkdir -p server/dist
@@ -188,7 +191,7 @@ else ifeq ($(OS),Linux)
 else ifeq ($(OS),Windows_NT)
 	cd server && env GOOS=windows GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-windows-amd64.exe;
 else
-	$(error make debug depends on uname to return your OS. If it does not return 'Darwin' (meaning OSX), 'Linux', or 'Windows_NT' (all recent versions of Windows), you will need to edit the Makefile for your own OS.)
+	$(error make server-debug depends on uname to return your OS. If it does not return 'Darwin' (meaning OSX), 'Linux', or 'Windows_NT' (all recent versions of Windows), you will need to edit the Makefile for your own OS.)
 endif
 
 	rm -rf dist/
@@ -201,13 +204,17 @@ endif
 # webapp-debug builds and deploys a debug version of the plugin's webapp
 .PHONY: webapp-debug
 webapp-debug:
+ifeq ($(wildcard ../mattermost-server/.*),)
+	$(error in order to use the make webapp-debug commands, your plugin's development directory needs to be located in the same parent directory as the mattermost-server (they need to be 'peers' or 'siblings').)
+endif
 
 ifneq ($(HAS_WEBAPP),)
 # link the webapp directory
+	rm -rf ../mattermost-server/plugins/$(PLUGIN_ID)/webapp
 	mkdir -p ../mattermost-server/plugins/$(PLUGIN_ID)/webapp
 	ln -nfs $(PWD)/webapp/dist ../mattermost-server/plugins/$(PLUGIN_ID)/webapp/dist
 # start an npm watch
-	cd webapp && $(NPM) run run &
+	cd webapp && $(NPM) run watch &
 endif
 
 	@echo "\n\n*** After the frontend is compiled, run 'make reset' to reset the plugin.\nRun 'make reset' every time a change is made to force the server to serve the chages in your webapp portion of the plugin.\nRun 'make stop' to stop the npm watcher.\n\n"
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ GO ?= $(shell command -v go 2> /dev/null)
 NPM ?= $(shell command -v npm 2> /dev/null)
 CURL ?= $(shell command -v curl 2> /dev/null)
 MANIFEST_FILE ?= plugin.json
+PWD := $(shell pwd)
+OS := $(shell uname 2> /dev/null)
 export GO111MODULE=on
 
 # You can include assets this directory into the bundle. This can be e.g. used to include profile pictures.
@@ -54,7 +56,7 @@ endif
 govet:
 ifneq ($(HAS_SERVER),)
 	@echo Running govet
-	@# Workaroung because you can't install binaries without adding them to go.mod 
+	@# Workaroung because you can't install binaries without adding them to go.mod
 	env GO111MODULE=off $(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
 	$(GO) vet ./server/...
 	$(GO) vet -vettool=$(GOPATH)/bin/shadow ./server/...
@@ -144,7 +146,7 @@ endif
 
 ## Creates a coverage report for the server code.
 .PHONY: coverage
-coverage: server/.depensure webapp/.npminstall
+coverage: webapp/.npminstall
 ifneq ($(HAS_SERVER),)
 	$(GO) test -race -coverprofile=server/coverage.txt ./server/...
 	$(GO) tool cover -html=server/coverage.txt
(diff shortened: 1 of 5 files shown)


crspeller:
Adding @lieut-data as he is the original author of the fancy deploy stuff. 

Also @hanzei as FYI.

debug-plugin.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
lieut-data: Worth putting this into `build/` or something vs. the top-level?

debug-plugin.sh
```diff
@@ -0,0 +1,20 @@
...
+
```
lieut-data: Suggest `set -euf -o pipefail` here to make the bash script safer.

debug-plugin.sh
```diff
@@ -0,0 +1,20 @@
...
+PLUGIN_ID=$(build/bin/manifest id)
+
+if [[ -z ${PLUGIN_ID} ]]
```
lieut-data: I /think/ `set -o pipefail` above might do this for free, though I'm not sure.

Makefile
```diff
@@ -167,3 +169,84 @@ endif
...
+	@echo "\n\n*** After the frontend is compiled, run 'make reset' to reset the plugin.\nRun 'make reset' every time a change is made to force the server to serve the chages in your webapp portion of the plugin.\nRun 'make stop' to stop the npm watcher.\n\n"
+
+# Reset the plugin
+.PHONY: reset
+reset:
+ifeq ($(and $(MM_SERVICESETTINGS_SITEURL),$(MM_ADMIN_USERNAME),$(MM_ADMIN_PASSWORD)),)
+	$(error In order to use make reset, the following environment variables need to be defined: MM_SERVICESETTINGS_SITEURL, MM_ADMIN_USERNAME, MM_ADMIN_PASSWORD)
+endif
+
+# If we were debugging, we have to unattach the delve process or else we can't disable the plugin.
```
lieut-data: Curious what the limitation is here -- why doesn't `delve` exit when the plugin process exits? Can it be made to do that?

Makefile
```diff
@@ -167,3 +169,84 @@ endif
...
+	mkdir -p ../mattermost-server/plugins/$(PLUGIN_ID)/webapp
+	ln -nfs $(PWD)/webapp/dist ../mattermost-server/plugins/$(PLUGIN_ID)/webapp/dist
+# start an npm watch
+	cd webapp && $(NPM) run run &
+endif
+
+	@echo "\n\n*** After the frontend is compiled, run 'make reset' to reset the plugin.\nRun 'make reset' every time a change is made to force the server to serve the chages in your webapp portion of the plugin.\nRun 'make stop' to stop the npm watcher.\n\n"
+
+# Reset the plugin
+.PHONY: reset
```
lieut-data: I'd love to see us eliminate the need for `make reset` by possibly only deploying via he API and perhaps not having to clean up delve processes as commented below.

Makefile
```diff
@@ -167,3 +169,84 @@ endif
...
+	$(eval TOKEN := $(shell curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep -o MMAUTHTOKEN=[0-9a-z]\* | cut -f2 -d'=' 2> /dev/null))
+	@curl -s -H "Authorization: Bearer $(TOKEN)" -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins/$(PLUGIN_ID)/disable > /dev/null && \
+		curl -s -H "Authorization: Bearer $(TOKEN)" -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins/$(PLUGIN_ID)/enable > /dev/null && \
+		echo "OK." || echo "Sorry, something went wrong. Check that MM_ADMIN_USERNAME and MM_ADMIN_PASSWORD env variables are set correctly."
+else
+	$(error In order to use make reset, you need to have curl installed.)
+endif
+
+# Stop the webpack
+.PHONY: stop
```
lieut-data: Instead of `stop` target, can we make `webapp-debug` blocking, and thus terminated with a SIGINT? This has long been a "mistake" in the mattermost-webapp, and something I hope we can eventually fix there as well.

Makefile
```diff
@@ -167,3 +169,84 @@ endif
...
+	cd server && env GOOS=windows GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-windows-amd64.exe;
+else
+	$(error make debug depends on uname to return your OS. If it does not return 'Darwin' (meaning OSX), 'Linux', or 'Windows_NT' (all recent versions of Windows), you will need to edit the Makefile for your own OS.)
+endif
+
+	rm -rf dist/
+	mkdir -p dist/$(PLUGIN_ID)/server/dist
+	cp $(MANIFEST_FILE) dist/$(PLUGIN_ID)/
+	cp -r server/dist/* dist/$(PLUGIN_ID)/server/dist/
+	mkdir -p ../mattermost-server/plugins
```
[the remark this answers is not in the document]
cpoile: The problem with using deploy semantics is the symlink to the dist folder... As far as I can tell, when a plugin is installed via the API, it calls os.RemoveAll(pluginPath).... Which would be fine, if we just went in and deleted the webapp/dist folder and relinked it. But then we'd have to rely on paths anyway.

If we want the webapp watcher benefits and the speed of just calling `make reset` for a webapp change, then I think we have to link the directories. So I'm okay with making it a requirement to develop the plugin in a peer directory. If someone doesn't want to, then they're free to put it wherever they want (they just can't use these make commands). 

I added a check and it will error out when its not a peer to the server. That should help new devs figure out what's wrong, and experienced devs who know what they're doing will decide if the tradeoff is worth it for them.

Makefile
```diff
@@ -167,3 +169,84 @@ endif
...
+	cp -r dist/* ../mattermost-server/plugins/
+
+# webapp-debug builds and deploys a debug version of the plugin's webapp
+.PHONY: webapp-debug
+webapp-debug:
+
+ifneq ($(HAS_WEBAPP),)
+# link the webapp directory
+	mkdir -p ../mattermost-server/plugins/$(PLUGIN_ID)/webapp
```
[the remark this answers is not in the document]
cpoile: I did a quick search and it seems anything like that is related to hooking into node, not really running shell commands. (at least what I could find)
For now I'm fine with pressing up and enter to do a `make reset` -- it's faster than what we're doing now. :)

webapp/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "This plugin serves as a starting point for writing a Mattermost plugin.",
   "main": "src/index.js",
   "scripts": {
+    "run": "webpack --progress --watch",
```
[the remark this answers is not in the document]
cpoile: Sure, I'm fine with that. Changed.

debug-plugin.sh
cpoile: It seems to belong in build, I agree. But I really like typing `./de`+`tab` and I don't want to have to type `./build/de`+`tab`. I'm lazy. :)

debug-plugin.sh
cpoile: great idea, done

debug-plugin.sh
cpoile: Yah, I'm not sure either... I'll leave it, unless you have a strong opinion (then I'll investigate and test).

Makefile
cpoile: From what I can tell, when it's in headless there's no convenient way to tell it to exit, and it won't let go of the process it's debugging (even if that process tries to exit).

Makefile
cpoile: Even leaving aside delve, we would still need to run `make reset`, because the mattermost server needs to reload the plugin's js before it can serve the changed files. But maybe I'm misunderstanding what happens when a plugin is loaded and served.

Makefile
cpoile: Personally I wouldn't want that -- I use the same terminal to run all these commands, with the webpack running in the background. I can see when the webpack is finished and then I can run the commands I need to. If it was blocking it would kind of negate the benefit of the always-on watching (for me at least)

debug-plugin.sh
cpoile: I did add a `kill-plugin.sh` to the build directory. I run that far less frequently, so it definitely belongs there. 
We will remove `kill-plugin.sh` when this bug of plugins not being shut down is taken care of.

cpoile:
@lieut-data lots of comments for you. thanks for the feedback!

debug-plugin.sh
lieut-data: It would be great to simplify if possible! (Reduces overall cognitive load when reading.)

Makefile
lieut-data: `make deploy` used to literally remove and then reinstall the plugin. That behaviour would trigger a websocket event to remove the code, followed by a websocket event to enable it, effectively forcing a client to refresh the code automatically.

I'm not sure of the semantics with the new `force=true` parameter -- @hanzei, do you know if that still triggers an uninstall/reinstall? If not, would it be easy to "fix" that?

Makefile
lieut-data: I can appreciate why it seems convenient at first, but there are a number of shortcomings:
* Anyone using a terminal editor (e.g. vim) will be spammed with logs from these background processes inside their editor.
* Running more than one background process makes it hard to decouple logs -- especially failure logs -- from a particular process, not to mention just using the console with something randomly deciding to write makes it hard to enter commands.
* We have often fielded questions from new contributors about why they can't start their server, only to realize they had already started it somewhere else but couldn't "find" it since it was in the background.

I'm 5/5 that we don't want to promulgate this behaviour into the sample plugin, especially when you can trivially opt-in with your own `make <target> &`.

As an aside, I'd strongly encourage the use of tmux or a similar equivalent and running your jobs in separate panes within the same window. At least, I find that way more powerful.

debug-plugin.sh
lieut-data: Oh, somehow I had assumed this script was going to be executed from the context of a `make debug` target. Actually, that seems slightly nicer since you can assume the proper input arguments and don't need to invoke `manifest id` yourself. 2/5 on keeping all the targets in the `Makefile` vs. having other entry points

Makefile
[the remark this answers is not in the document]
lieut-data: Right, but you can invoke any shell commands you need from `node`. Thinking something like https://www.npmjs.com/package/on-build-webpack that kicks off a `make deploy` or whatever.

Pushing back on this mostly because this isn't just code for us to use temporarily -- this is /the/ example for how to develop Mattermost plugins. People will fork this repository and likely be stuck with whatever tooling we give them for some time (not everyone updates against mattermost-plugin-sample regularly).

It would be terrific if `make webapp-debug` launched the webpack process, listened for compilation events, and effectively triggered a `BUILD_FAST=1 make deploy` that, in turn, reloaded the plugin client-side. You wouldn't even need to reload your browser in that case :)

Makefile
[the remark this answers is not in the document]
hanzei: Maybe we can offer a environment variable that devs have to set similar to what we do with `MM_SERVICESETTINGS_SITEURL`? The default would be look for the server at ` ../mattermost-server/`, but if the the environment variable is set, we use this instead.

Makefile
hanzei: Installing a plugin over the API does a few things:
- Kill old plugin process
- Start a new plugin process
- Reload the js bundle

I'm not sure what the goal of 'reset` is. Just to deal with `delve`?

Makefile
hanzei: Strong :+1: for what Jesse said.

debug-plugin.sh
hanzei: Agree that this should be part of the Makefile.

build/kill-plugin.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
hanzei: Why is this script needed? This seams to be a task for the server to take care of.

cpoile:
Will reopen once we've had time to use https://github.com/mattermost/mattermost-plugin-jira/pull/255.

PR #43, opened 2019-06-19
Refactoring

hanzei:
This PR contains to much (Sorry, dear reviewers, I will do better next time) It:
- Cleans up legacy files
- Adds a `min_server_version` to the manifest
- Changes to the code to be golint compliant
- Update dependencies to MM 5.12
- Downstreams `i18n-extract` from the demo plugin

The changes can be best reviewed as individual commits.


plugin.json
```diff
@@ -3,6 +3,7 @@
     "name": "Sample Plugin",
     "description": "This plugin serves as a starting point for writing a Mattermost plugin.",
     "version": "0.1.0",
+    "min_server_version": "5.2.0",
```
lieut-data: nit: Did 5.2.0 actually support `min_server_version`? Also, if we're depending on 5.12.0 now, should this just get bumped to 5.12.0, and maybe we tag commits supporting an older version?

server/plugin.go
```diff
@@ -19,8 +20,9 @@ type Plugin struct {
 	configuration *configuration
 }
 
+// This example demonstrates a plugin that handles HTTP requests which respond by greeting the world.
```
lieut-data: Can we also become godoc compliant by writing this as follows?
```suggestion
ServeHTTP demonstrates a plugin that handles HTTP requests by greeting the world.
```

server/plugin.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/mattermost/mattermost-server/plugin"
 )
 
+// Plugin is the object to run the plugin
```
lieut-data: nit: I'm not fond of the phrase `object to run the plugin`. Perhaps something like:
```suggestion
// Plugin implements the interface expected by the Mattermost server to communicate between the server and plugin processes.
```

Makefile
```diff
@@ -150,6 +152,17 @@ ifneq ($(HAS_SERVER),)
 	$(GO) tool cover -html=server/coverage.txt
 endif
 
+## Extract strings for translation from the source code.
+.PHONY: i18n-extract
+i18n-extract:
+ifneq ($(HAS_WEBAPP),)
+ifeq ($(HAS_MM_UTILITIES),)
+	@echo "You must clone github.com/mattermost/mattermost-utilities repo in .. to use this command"
```
lieut-data: Out of scope for this PR, but I wonder if we could go get this instead, or perhaps build it as a dependency into the `manifest` tool somehow. Also, I guess we don't support translating strings embedded in the Go source?

lieut-data:
Oh, and can we update the `Makefile` target to install / run `golint`?

plugin.json
hanzei: > Did 5.2.0 actually support `min_server_version'.

No. `min_server_version` was added in 5.6.0

> Also, if we're depending on 5.12.0 now, should this just get bumped to 5.12.0, and maybe we tag commits supporting an older version?

I can see four different version we can require here:
1. 5.2.0: This is "real" required server version, as this is required to run the plugin.
2. 5.6.0: Because this added `min_server_version` and the setting will only be taken into account from the version on.
3. 5.10.0: Because this added e.g. asset handling, which we support in the Makefile
4. 5.12.0: Because this is the latest MM version and we "build" on it. And it added support for public files, which we support in the Makefile.

I'm happy to bump this to 5.12 if there a no concerns that this will "limit" plugin development for older servers. Let me know what do you think about this.

server/plugin.go
hanzei: I'm a bit surprised `golint ./...` didn't caught this one.

Makefile
hanzei: > Out of scope for this PR, but I wonder if we could go get this instead, or perhaps build it as a dependency into the `manifest` tool somehow. 

Maybe we can do this. But to me translating `webapp` is a rare case. And installing `mattermost-utilities` not to big of a task. Feel free to open an issue for this.

> Also, I guess we don't support translating strings embedded in the Go source?

`i18n-extract` only extracts the strings in `webapp` ATM. There is an ongoing discussion about how server translations should work. See https://github.com/mattermost/mattermost-server/pull/10949.

hanzei:
> Oh, and can we update the `Makefile` target to install / run `golint`?

:+1: 
Done

hanzei:
Switching to @marianunez as a reviewer because of the changed responsibilities.

plugin.json
lieut-data: My gut instinct is to bump this to match `go.mod` and thus `5.12.0`, and iterate separately on how to enable plugin development for older server versions. It might just amount to an FAQ in the README.md describing how to (optionally?) downgrade the `go.mod` dependency and `min_sever_version`, and describing the caveats accordingly.

Makefile
```diff
@@ -36,7 +36,7 @@ endif
 gofmt:
 ifneq ($(HAS_SERVER),)
 	@echo Running gofmt
-	@for package in $$(go list ./server/...); do \
```
lieut-data: Hmm, do we need to drop the `server` prefix here? This only gets run if `HAS_SERVER`, so it won't help catch linting issues say, in the manifest tool, if it's a webapp only plugin anyway. 0/5, since it probably doesn't hurt.

hanzei: Good catch, that this only get run if `HAS_SERVER`. My idea was that this was checking the files in `build` "for us" and catch issues when we change these files.

plugin.json
hanzei: I'm fine with your proposed approach. Would you take care of creating a ticket for the documentation or do the changes yourself?

PR #44, opened 2019-06-19
Clarify Go modules requirements.

lieut-data:
For anyone new to Go modules, it's worth adding some clarity on how to use this project in the event it remains located in the `$GOPATH`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,7 @@ Alternatively shallow clone the repository to a directory outside of `$GOPATH` m
 git clone --depth 1 https://github.com/mattermost/mattermost-plugin-sample com.example.my-plugin
 ```
 
-<aside class="notice">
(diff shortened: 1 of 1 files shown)


ThiefMaster:
The `<aside>` seems to break markdown rendering and not actually affect styling:

https://github.com/mattermost/mattermost-plugin-sample/blob/34afd70a5d234cd7f8b12d578ea5bb08b09277d1/README.md

ali-farooq0:
@ThiefMaster interesting, what's the syntax to show a warning/note in MD? Just found the following snippet, not use if this is the convention though.
```
> **Note:** This is  a note.
```

> **Note:** This is  a note.

ThiefMaster:
I don't think there is really any note syntax in markdown. `> ...` is generally meant for quotes.

ali-farooq0:
Well then ;p

hanzei:
What is the downside of just using "normal" md text?

lieut-data:
Huh, weird. I swear I came across this in GitHub documentation, and figured the rendering was a pre-commit issue, but you're right this doesn't seem to work at all. My hope was to draw attention to this section, but I suspect that's just overkill. Will fix!

PR #45, opened 2019-06-21
Add support for custom makefile

hanzei:
This PR adds support for custom makefiles. The custom makefile must be located under `build/custom.mk`. I don't see the need for making this customisable. I've added a empty file as a boiler plate.

Please note that custom targets will **not** be listed via `make help`. I tried fixing this but it took to much time. I would open a Help Wanted ticket for this, if there are no objections.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ include build/setup.mk
 
 BUNDLE_NAME ?= $(PLUGIN_ID)-$(PLUGIN_VERSION).tar.gz
 
+ifneq ($(wildcard build/custom.mk),)
+	include build/custom.mk
+endif
+
 ## Checks the code style, tests, builds and bundles the plugin.
 all: check-style test dist
 
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ include build/setup.mk
 
 BUNDLE_NAME ?= $(PLUGIN_ID)-$(PLUGIN_VERSION).tar.gz
 
+# Include custom makefile, if pressent
 ifneq ($(wildcard build/custom.mk),)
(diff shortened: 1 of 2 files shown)


ali-farooq0:
Thanks @hanzei! Was just wondering what's the advantage of adding a boiler plate custom make file? Is it to set a pattern for whoever uses this as a starting point build a plugin?

Makefile
```diff
@@ -12,6 +12,10 @@ include build/setup.mk
 
 BUNDLE_NAME ?= $(PLUGIN_ID)-$(PLUGIN_VERSION).tar.gz
 
+ifneq ($(wildcard build/custom.mk),)
+	include build/custom.mk
```
marianunez: @hanzei What type of custom commands would be included in this makefile? I see that each plugin has their own build/setup.mk that is already included in Makefile. Could these custom commands be added there ? 🤔

hanzei:
Hey @marianunez and @ali-farooq0,

Sorry, I didn't made my intentions clear. Let me make up for this:

Plugin devs start of with a new plugin my copying this repo. The make changes to the go and react code base. Concurrently, this repository gets updated with e.g. new `make` targets. We currently don't have any tooling in place to update the Makfile. Hence, the only strategy is copy&past the update makefile from the repo. To make process more comfortable there should be "no merge conflicts". Ideally plugin devs should have no reason to "fork" `Makefile`. To accomplish this I want to encourage devs to move the custom changes like new targets or changed environment variables in this custom makefile.

Let me know what you think about this.

hanzei:
Thanks for the feedback. I added a short comment about this.

PR #46, opened 2019-06-21
Make go flags customisable

hanzei:
I've run into cases where I wanted to customise the go flags and thought this might be useful in this file.

I opted to have two different variables because there are some flags just for one command.


Makefile
```diff
@@ -2,6 +2,8 @@ GO ?= $(shell command -v go 2> /dev/null)
 NPM ?= $(shell command -v npm 2> /dev/null)
 CURL ?= $(shell command -v curl 2> /dev/null)
 MANIFEST_FILE ?= plugin.json
+GO_TEST_FLAGS ?= -race
+GO_BUILD_FLAGS ?=
```
hanzei: Added this here to make clear what the comment of the variable is. 1/5

Makefile
```diff
@@ -136,7 +138,7 @@ endif
 .PHONY: test
 test: webapp/.npminstall
 ifneq ($(HAS_SERVER),)
-	$(GO) test -race -v ./server/...
```
hanzei: I intentionally removed `-v` here. `-v` is handy with a low number of tests, but with larger plugin it just spams the log. 2/5

OT: Maybe we can add support for `make test -v`.

Makefile
ali-farooq0: sorry, to make clear which comment?

hanzei: Dump type. Please take another look.

ali-farooq0:
Looks like this has a conflict now :(

Makefile
hanzei: > imagining a CI pipeline as the primary client

The usually case I had in mind was a developer that is testing his/her changes before committing. But the CI case convinced me.

PR #51, opened 2019-07-02
Add a task of webapp test

kaakaa:
Add a task for running webapp test.


hanzei:
@marianunez Yes

webapp/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "webpack --mode=production",
     "lint": "eslint --ignore-pattern node_modules --ignore-pattern dist --ext .js --ext .jsx . --quiet",
-    "fix": "eslint --ignore-pattern node_modules --ignore-pattern dist --ext .js --ext .jsx . --quiet --fix"
+    "fix": "eslint --ignore-pattern node_modules --ignore-pattern dist --ext .js --ext .jsx . --quiet --fix",
+    "test": "echo \"no test specified\""
```
marianunez: Should we complete this test script definition with the current setup with have with jest in the webapp? I'm wondering if it will be better to leave that as a boilerplate over this message  🤔

lieut-data: Agree with @marianunez -- I think we need this to be more fleshout, otherwise we're just kicking the problem down into the stack.

@kaakaa, would you be open to:
1) Updating the package dependencies to rely on jest, and changing this target to run jest appropriately?
2) Add a very simple test that could be used as a starting point for plugin authors?

kaakaa:
Thank you for commenting @marianunez and @lieut-data.

I've fleshed out tests with [hello-world plugin](https://developers.mattermost.com/extend/plugins/webapp/hello-world/#setting-up-the-workspace) from developers document. Is this too much for template?

webapp/src/components/icon.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export default class Icon extends React.PureComponent {
+    render() {
+        return (
+            <i className='icon fa fa-plug'/>
+        );
+    }
+}
```
lieut-data: nit: missing newline here

kaakaa:
Thanks @lieut-data!
It sounds good to make test cases more simple, and I've simplify it.

> there's also a number of proposed dependency updates here. 

I supposed that these updates are needed for running Jest. But I'm not familiar with Jest, so if there are any ways for running Jest without updates, please let me know it.

webapp/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+    presets: ['@babel/preset-env', '@babel/preset-react']
+}
```
lieut-data: There's a missing trailing newline. I'd suggest installing the appropriate editorconfig.org plugin for your editor to prevent this. Missing this means future readers (who have their editors appropriately configured) who open the file are likely to force an unexpected change to the file.

webapp/package.json
```diff
@@ -6,22 +6,24 @@
...
     "lint": "eslint --ignore-pattern node_modules --ignore-pattern dist --ext .js --ext .jsx . --quiet",
-    "fix": "eslint --ignore-pattern node_modules --ignore-pattern dist --ext .js --ext .jsx . --quiet --fix"
+    "fix": "eslint --ignore-pattern node_modules --ignore-pattern dist --ext .js --ext .jsx . --quiet --fix",
+    "test": "jest --forceExit --detectOpenHandles --verbose"
   },
   "author": "",
   "license": "",
   "devDependencies": {
-    "babel-core": "^6.26.3",
+    "@babel/core": "^7.5.0",
```
crspeller: Can you remove the unrelated dependency updates? We should handle these in a separate PR pulling in my changes from: https://github.com/mattermost/mattermost-plugin-demo/pull/42

lieut-data: Aha, that's where they came from! Thanks, @crspeller -- @kaakaa, I can get these merged to `mattermost-plugin-starter-template`, and then a merge from upstream should resolve this concern.

lieut-data: https://github.com/mattermost/mattermost-plugin-starter-template/pull/53 is up for review.

kaakaa: Thanks @crspeller and @lieut-data.

I've merged it from upstream 👍

kaakaa:
Thank you for reviewing @crspeller and @lieut-data!
And sorry for my careless force push. I'll keep it in mind.

I've fixed the points that you reviewed, so please take a look again.

PR #56, opened 2019-08-09
GH-10166 - Call to CheckRequiredConfig when OnActivate is hit

g3rv4:
#### Summary
Show how to use `CheckRequiredConfig` at `OnActivate`.

Needs the changes on the server merged before this can be merged, as the new API method would not be available otherwise.

#### Ticket Link
[#10166 - Plugin framework: Show that a plugin requires a certain Mattermost config setting](https://github.com/mattermost/mattermost-server/issues/10166)

#### Related PRs
Server change: [Plugin framework: Show that a plugin requires a certain Mattermost config setting ](https://github.com/mattermost/mattermost-server/pull/11832)

### DIFF
--- a/build/manifest/main.go
+++ b/build/manifest/main.go
@@ -12,12 +12,20 @@ import (
 
 const pluginIDGoFileTemplate = `package main
 
+import (
+	"strings"
+
+	"github.com/mattermost/mattermost-server/model"
+)
+
 var manifest = struct {
-	ID      string
-	Version string
+	ID             string
+	Version        string
+	RequiredConfig *model.Config
 }{
-	ID:      "%s",
-	Version: "%s",
+	ID:             "%s",
+	Version:        "%s",
+	RequiredConfig: model.ConfigFromJson(strings.NewReader(` + "`%s`" + `)),
 }
 `
 
@@ -101,7 +109,7 @@ func applyManifest(manifest *model.Manifest) error {
 	if manifest.HasServer() {
 		if err := ioutil.WriteFile(
 			"server/manifest.go",
-			[]byte(fmt.Sprintf(pluginIDGoFileTemplate, manifest.Id, manifest.Version)),
+			[]byte(fmt.Sprintf(pluginIDGoFileTemplate, manifest.Id, manifest.Version, model.ConfigToJsonWithoutEmptyFields(manifest.RequiredConfig))),
 			0644,
 		); err != nil {
 			return errors.Wrap(err, "failed to write server/manifest.go")
--- a/plugin.json
+++ b/plugin.json
@@ -18,5 +18,10 @@
         "header": "",
         "footer": "",
         "settings": []
+    },
+    "requires_config": {
+        "ServiceSettings": {
+            "EnablePostUsernameOverride": true
+        }
     }
 }
--- a/plugin.json
+++ b/plugin.json
@@ -19,7 +19,7 @@
         "footer": "",
         "settings": []
     },
-    "requires_config": {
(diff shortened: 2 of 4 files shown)


plugin.json
```diff
@@ -18,5 +18,10 @@
         "header": "",
         "footer": "",
         "settings": []
+    },
+    "requires_config": {
+        "ServiceSettings": {
```
g3rv4: Do we want to include a requirement on the sample? I'd say YES if we're showing how `CheckRequiredConfig` should be used

hanzei:
Deferring a review until the server PR is merged.

build/manifest/main.go
```diff
@@ -101,7 +109,7 @@ func applyManifest(manifest *model.Manifest) error {
 	if manifest.HasServer() {
 		if err := ioutil.WriteFile(
 			"server/manifest.go",
-			[]byte(fmt.Sprintf(pluginIDGoFileTemplate, manifest.Id, manifest.Version)),
+			[]byte(fmt.Sprintf(pluginIDGoFileTemplate, manifest.Id, manifest.Version, model.ConfigToJsonWithoutEmptyFields(manifest.RequiredConfig))),
```
lieut-data: Oh, I didn't realize we were using `ConfigToJsonWithoutEmptyFields` here as well. Comment added to original PR, but I'm wondering if we could just skip the call and let `model.ConfigFromJson` receive all the `nil` fields -- does that hurt anything?

plugin.json
lieut-data: Actually, ideally we'd put the "demo" in our https://github.com/mattermost/mattermost-plugin-demo, and just leave `required_config` as an empty struct here. The intent of the starter plugin is to provide the minimal boilerplate to get started, and adding a demo in this case makes it one more thing to "remove" to actually get started.

server/activate.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/mattermost/mattermost-server/model"
+)
+
+// OnActivate is executed when the plugin is activated
+func (p *Plugin) OnActivate() error {
+	p.API.LogDebug("Activating plugin")
```
lieut-data: nit: let's remove this

server/activate.go
```diff
@@ -0,0 +1,26 @@
...
+
+	isCompatible, err := p.Helpers.CheckRequiredConfig(manifest.RequiredConfig, p.API.GetConfig())
```
lieut-data: See comment in server PR, but the helpers have access to `p.API` directly so no need to pass the actual config. Seeing the example code here, I'm also wondering if we should have the `CheckRequiredConfig` return an error when the config is incompatible, allowing it to internalize the description of that error instead of forcing every plugin to do lines 19-23 for themselves.

Thoughts? It would return an error both when it failed to check, and when the config mismatched.

server/manifest.go
```diff
@@ -1,9 +1,17 @@
...
-	Version string
+	ID             string
+	Version        string
+	RequiredConfig *model.Config
 }{
-	ID:      "com.mattermost.plugin-starter-template",
-	Version: "0.1.0",
+	ID:             "com.mattermost.plugin-starter-template",
+	Version:        "0.1.0",
+	RequiredConfig: model.ConfigFromJson(strings.NewReader(`{"ServiceSettings":{"EnablePostUsernameOverride":true}}`)),
```
lieut-data: I know we're generating this, but I wonder if there's anyway to generate the literal `&model.Config` instead.

PR #57, opened 2019-08-14
Mark manifest.[go|js] as a generated file

ethervoid:
Fixes https://github.com/mattermost/mattermost-server/issues/11862

Now the autogenerated manifest files have:

- Comment showing that the file is generated and you should not edit manually
- .gitattributes that:
  - avoids to show the diff in the console git command for those files
  - takes those files as binaries when a merge is done
  - collapse those file diffs in github, are treated as binaries.


build/manifest/main.go
```diff
@@ -10,7 +10,11 @@ import (
 	"github.com/pkg/errors"
 )
 
-const pluginIDGoFileTemplate = `package main
+const autogeneratedFileComment = "// This file is automatically generated. Do not modify it manually."
```
lieut-data: It would be duplicated, but I'm could we just embed this comment into the `pluginIDGoFileTemplate` directly? I think it would read more cleanly.

webapp/src/manifest.js
```diff
@@ -1,2 +1,3 @@
+// This file is automatically generated. Do not modify it manually.
```
lieut-data: nit: perhaps add a newline after this to match `manifest.go`?

.gitattributes
```diff
@@ -0,0 +1,4 @@
+manifest.go -diff -merge
```
lieut-data: Can we specify the paths to these files so as to avoid matching on potentially other files that get added to the repository with the same names but different purposes?

ethervoid:
@lieut-data requested changes made :)

.gitattributes
```diff
@@ -0,0 +1,4 @@
+server/manifest.go -diff -merge
```
hanzei: I'm not convinced that this line is needed. These files aren't binary files and can be easily checked. Why do opted to add these line?

ethervoid: My idea with those lines was:

- avoid to show the diff in the cli to reduce the noise
- treat it as a binary, because is a generated file, so we can avoid merge conflicts

I thought that it could be useful but if ypu think is not I can remove then :)

hanzei: I'm 1/5 that I actually want to see the diff on the cli. This allows me to double check that the version bump was made correctly.

I see your point on 2., but I suppose there are rarely merge conflicts and if so I want to resolve them myself.

@lieut-data Thoughts on the discussion?

lieut-data: I'm 0/5 right now. I agree that it might be "too magical" to hide in the CLI.

hanzei: @ethervoid Would you be fine with removing these lines?

ethervoid: Sure! Done :)

hanzei:
Thanks @ethervoid :+1:

PR #64, opened 2019-09-01
add typescript support

niklabh:
#### Summary
- Add typescript support for writing files with ts and tsx extensions

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/12019

https://mattermost.atlassian.net/browse/MM-18155

### DIFF
--- a/webapp/.eslintrc.json
+++ b/webapp/.eslintrc.json
@@ -632,5 +632,47 @@
         "onlyEquality": false
       }
     ]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["**/*.tsx", "**/*.ts"],
+      "extends": [
+        "./node_modules/eslint-config-mattermost/.eslintrc.json",
+        "./node_modules/eslint-config-mattermost/.eslintrc-react.json",
+        "plugin:@typescript-eslint/recommended",
+        "plugin:cypress/recommended"
+      ],
+      "rules": {
+        "@typescript-eslint/no-var-requires": 0,
+        "@typescript-eslint/prefer-interface": 0,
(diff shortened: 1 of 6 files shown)


webapp/.eslintrc.json
```diff
@@ -632,5 +632,47 @@
         "onlyEquality": false
       }
     ]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["**/*.tsx", "**/*.ts"],
+      "extends": [
```
jespino: This is not working properly, but I don't know why is not failing in mattermost-webapp. I'm investigating it.

jespino: My fault I hadn't upgrade to eslint 6 in my repo.

webapp/.eslintrc.json
```diff
@@ -632,5 +632,47 @@
...
+        "./node_modules/eslint-config-mattermost/.eslintrc.json",
```
jespino: I think this shouldn't be here. We are extending the previous existing eslint config, so here we only should add the `"plugin:@typescript-eslint/recommended"` line

webapp/.eslintrc.json
jespino: Here we can replace it directly with the typescript plugin setting.
```suggestion
      "extends": "plugin:@typescript-eslint/recommended",
```

webapp/.eslintrc.json
niklabh: removed

webapp/.eslintrc.json
niklabh: done

PR #70, opened 2019-11-04
fix for go 1.13 strictness

cpoile:
I would love to know if there's a better fix than doing this for every plugin repo. But I guess not, since it's a bad import in server@5.12.  @hanzei thoughts?


cpoile:
@hanzei Not exactly, try changing it to `5.14.0+incompatible`, `5.16.0+incompatible` and `5.16.0` and see the errors. Things are a little messed up somewhere. @levb and I tried to figure it out last week, but didn't really get anywhere.

hanzei:
Could you please double check? Since MM 5.16 we are using the latest version of `github.com/willnorris/imageproxy` with a correct date in the manifest.

cpoile:
@hanzei It's other things -- go can't even find `5.16.0+incompatible`. That's why I'm suggesting you to try it out yourself, the errors are weird.

ali-farooq0:
@cpoile @hanzei is this still applicable or should this be closed?

hanzei:
We should bump the depencie to 5.16.X and use the hash.

plugin.json
```diff
@@ -3,7 +3,7 @@
     "name": "Plugin Starter Template",
     "description": "This plugin serves as a starting point for writing a Mattermost plugin.",
     "version": "0.1.0",
-    "min_server_version": "5.12.0",
+    "min_server_version": "5.16.0",
```
hanzei: I don't think we need to bump this. You can use 5.16 dependencies with 5.12

Issue #69, opened 2019-11-01
Fix `make deploy` to correctly report an error


lieut-data:
Running `make deploy` can appear to succeed:

```
> make deploy
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   735  100   679  100    56   7216    595 --:--:-- --:--:-- --:--:--  7223
Installing plugin via API
OK.
```

when in fact the configured `MM_ADMIN_USERNAME` or `MM_ADMIN_PASSWORD` was rejected by the server:
```
DEBU[2019-11-01T15:15:56.513026-04:00] Received HTTP request                         caller="web/handlers.go:77" method=POST request_id=97xj8ku85i81u89mq3oudjxf9c url=/api/v4/plugins/com.mattermost.webex/
enable
DEBU[2019-11-01T15:15:56.5131168-04:00] Invalid or expired session, please login again.  caller="mlog/log.go:162" err_details=UserRequired err_where= http_code=401 ip_addr="::1" method=POST path=/api/v4/p
lugins/com.mattermost.plugin-starter-template/enable request_id=97xj8ku85i81u89mq3oudjxf9c user_id=
```

The issue is that a login failure doesn't propagate outside of the attempt to capture the resulting token:
https://github.com/mattermost/mattermost-plugin-starter-template/blob/69b6ae8467fc7e80d7f128556e28de292fc289ab/Makefile#L148

Find a way to continue capturing the token but detect a login error (or any other upload issue), and ideally without using temporary files.

----
If you're interested please comment here and come [join our "Contributors" community channel](https://community.mattermost.com/core/channels/tickets) on our daily build server, where you can discuss questions with community members and the Mattermost core team. For technical advice or questions, please  [join our "Developers" community channel](https://community.mattermost.com/core/channels/developers).

New contributors please see our [Developer's Guide](https://developers.mattermost.com/contribute/getting-started/).

nathanmkaya:
Hey @lieut-data can I have a go at this?

hanzei:
Sure, thanks @nathanmkaya :+1:

PR #71, opened 2019-11-04
Fix Issue #69 

nathanmkaya:
Added a conditional of the token value and returns error if value is nil

Fixes: https://github.com/mattermost/mattermost-plugin-starter-template/issues/69

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ deploy: dist
 ## or copying the files directly to a sibling mattermost-server directory.
 ifneq ($(and $(MM_SERVICESETTINGS_SITEURL),$(MM_ADMIN_USERNAME),$(MM_ADMIN_PASSWORD),$(CURL)),)
 	@echo "Installing plugin via API"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -145,7 +145,10 @@ deploy: dist
 ## or copying the files directly to a sibling mattermost-server directory.
 ifneq ($(and $(MM_SERVICESETTINGS_SITEURL),$(MM_ADMIN_USERNAME),$(MM_ADMIN_PASSWORD),$(CURL)),)
 	@echo "Installing plugin via API"
-	$(eval TOKEN := $(shell curl -i --post301 --location $(MM_SERVICESETTINGS_SITEURL) -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep Token | cut -f2 -d' ' 2> /dev/null))
+	$(eval TOKEN := $(shell curl -i --fail --post301 --location $(MM_SERVICESETTINGS_SITEURL) -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep Token | cut -f2 -d' ' 2> /dev/null))
+	@if [$(TOKEN) == '']; then\
+		$(error Could not create token)\
```
lieut-data: If it’s not possible to extract the actual error, might we clarify the error message here, perhaps to say, “Failed to authenticate and extract token using $(MM_ADMIN_USERNAME): check server logs.”

nathanmkaya: Hey @lieut-data I Have added a more clear error message verbatim what you had suggested

PR #74, opened 2019-11-16
make deploy reliable using Go

lieut-data:
A recent change to `make deploy` broke the ability to actually deploy the plugin successfully. Let's fix this once and for all by moving to a Go binary that can use the `Client4` API properly.

Relates-to: https://github.com/mattermost/mattermost-plugin-starter-template/issues/69.
Fixes: https://mattermost.atlassian.net/browse/MM-20610

### DIFF
--- a/build/deploy/main.go
+++ b/build/deploy/main.go
@@ -16,7 +16,10 @@ import (
 func main() {
 	err := deploy()
 	if err != nil {
-		log.Print(err.Error())
+		fmt.Printf("Failed to deploy: %s\n", err.Error())
+		fmt.Println()
+		fmt.Println("Usage:")
+		fmt.Println("    deploy <plugin id> <bundle path>")
 		os.Exit(1)
 	}
 }
@@ -24,10 +27,10 @@ func main() {
 // deploy handles deployment of the plugin to a development server.
 func deploy() error {
 	if len(os.Args) < 3 {
-		return errors.New("deploy <plugin id> <bundle path>")
+		return errors.New("invalid number of arguments")
 	}
 
-	pluginId := os.Args[1]
+	pluginID := os.Args[1]
 	bundlePath := os.Args[2]
 
 	siteURL := os.Getenv("MM_SERVICESETTINGS_SITEURL")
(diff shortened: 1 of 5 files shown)


build/deploy/main.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+// deploy handles deployment of the plugin to a development server.
+func deploy() error {
+	if len(os.Args) < 3 {
+		return errors.New("deploy <plugin id> <bundle path>")
+	}
+
+	pluginId := os.Args[1]
```
hanzei: Given that this is freshly written code and we are encouraging plugin authors to write golint compliant code, I would like to see us following the convention here and writing golint compliant code our self.

build/deploy/main.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/mattermost/mattermost-server/model"
+	"github.com/mholt/archiver/v3"
+	"github.com/pkg/errors"
+)
+
+func main() {
+	err := deploy()
+	if err != nil {
+		log.Print(err.Error())
```
hanzei: Maybe we change this to 
```suggestion
		log.Printf("Failed to deploy plugin: %s" + perr.Error())
```
and modify the other error messages to start with a lower case latter? This would be golint compliant.

lieut-data:
@hanzei, can you prioritize this review? I'd like to get this merged before folks start hacking on plugins this Friday.

@prapti, I know there are a number of ongoing testing efforts right now, but I'm wondering if you might also be able to help prioritize this review? I'm also willing to take responsibility for the changes here, and since we don't "deploy" this code, it won't have any immediate impact on the product.

prapti:
@lieut-data The changes look good to me, however the last info logged with `deploy` is `Server restart and manual plugin enabling required` which seems a bit misleading since the plugin is installed successfully and shows up in Marketplace as well as System Console list on a refresh.

lieut-data:
Thanks, @prapti! That log /should/ only be emitted if the installation was done by copying to the sibling `mattermost-server` directory, which does indeed require a restart to show /that/ version of the plugin. Would it make sense to update the message to say the following instead?
```
	log.Print("Server restart required to load updated plugin.")
```

prapti:
> if the installation was done by copying to the sibling `mattermost-server` directory

@lieut-data Ah! That is indeed how I just tested. But, if I understood it correctly, if I upgraded to a newer version manually (say from 0.1.0 to 0.2.0), I should require a server restart for the plugin to display the correct 0.2.0 version number in System Console (and Marketplace)? Because right now, a mere refresh is doing that just fine, unless I'm missing a point here.

lieut-data:
@prapti, ah, fascinating observation! I can reproduce what you're seeing, but in fact the plugin is not updated -- if you check for any changed behaviour, it won't actually be present until server restart. (I tested this by modifying the starter template as such.)

The issue seems to be that the plugin management page -- to my surprise -- reports on the plugin version /found on disk/ instead of what's actually loaded into memory. I think it does this so it can report on plugins that aren't actually activated, but it doesn't overlay with data from what's actively running. I've filed https://mattermost.atlassian.net/browse/MM-20619 to address, but looking at the code, it's definitely been this way for a long time.

prapti:
@lieut-data Ha! So that is a bug albeit unrelated to this PR. Thank you for reporting :)
In that case, your suggestion to log `"Server restart required to load updated plugin."` instead of what's there currently makes total sense. I think it'll be less confusing.

lieut-data:
@prapti: done! Thanks for the review :)

lieut-data:
@marianunez, if you have a chance to take a peek at this, I think it will unblock hackathon efforts later this week. Adding you in case @hanzei doesn't have a chance to provide the second review, though I note I've addressed his feedback above.

PR #75, opened 2019-11-26
npm audit fix --save

lieut-data:
This addresses security fixes with an `npm audit fix --save`.


lieut-data:
Oops, just realized this updated the dependencies past what's in `mattermost-webapp`. Revisiting.

PR #79, opened 2020-01-09
Allow make deploy using token

lieut-data:
#### Summary
Introduce support for `MM_ADMIN_TOKEN` as an alternative to defining `MM_ADMIN_USERNAME` and `MM_ADMIN_PASSWORD`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,18 +34,18 @@ This will produce a single plugin file (with support for multiple architectures)
 dist/com.example.my-plugin.tar.gz
 ```
 
-There is a build target to automate deploying and enabling the plugin to your server, but it requires configuration of a [personal access token](https://docs.mattermost.com/developer/personal-access-tokens.html):
-```
-export MM_SERVICESETTINGS_SITEURL=http://localhost:8065
-export MM_ADMIN_TOKEN=j44acwd8obn78cdcx7koid4jkr
-make deploy
-```
-
-or login credentials:
+There is a build target to automate deploying and enabling the plugin to your server, but it requires login credentials:
 ```
 export MM_SERVICESETTINGS_SITEURL=http://localhost:8065
 export MM_ADMIN_USERNAME=admin
 export MM_ADMIN_PASSWORD=password
+make deploy
+```
+
+or configuration of a [personal access token](https://docs.mattermost.com/developer/personal-access-tokens.html):
+```
+export MM_SERVICESETTINGS_SITEURL=http://localhost:8065
+export MM_ADMIN_TOKEN=j44acwd8obn78cdcx7koid4jkr
 make deploy
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -34,7 +34,14 @@ This will produce a single plugin file (with support for multiple architectures)
...
 
-There is a build target to automate deploying and enabling the plugin to your server, but it requires configuration and [http](https://httpie.org/) to be installed:
+There is a build target to automate deploying and enabling the plugin to your server, but it requires configuration of a [personal access token](https://docs.mattermost.com/developer/personal-access-tokens.html):
+```
+export MM_SERVICESETTINGS_SITEURL=http://localhost:8065
+export MM_ADMIN_TOKEN=j44acwd8obn78cdcx7koid4jkr
+make deploy
+```
+
+or login credentials:
```
hanzei: What do you think about reversing the order? Deploying the plugin to a local instance is the usual case and a access token is not as handy, as it gets deleted on a server nuke.

build/deploy/main.go
```diff
@@ -34,11 +34,29 @@ func deploy() error {
...
+
+	if siteURL != "" {
+		client := model.NewAPIv4Client(siteURL)
+
+		if adminToken != "" {
+			log.Printf("Authenticating using token against %s.", siteURL)
+			client.SetToken(adminToken)
+
+			return uploadPlugin(client, pluginID, bundlePath)
+		} else if adminUsername != "" && adminPassword != "" {
```
hanzei: No need for the `else`

build/deploy/main.go
```diff
@@ -34,11 +34,29 @@ func deploy() error {
...
+			client := model.NewAPIv4Client(siteURL)
+			log.Printf("Authenticating as %s against %s.", adminUsername, siteURL)
+			_, resp := client.Login(adminUsername, adminPassword)
+			if resp.Error != nil {
+				return fmt.Errorf("Failed to login as %s: %s", adminUsername, resp.Error.Error())
```
hanzei: We should wrap the error
```suggestion
				return errors.Wrap(err, "Failed to login as %", adminUsername)
```

build/deploy/main.go
lieut-data: This is catching the case where just `siteURL` is defined, but no useful credentials, where we continue to fall back to the copy target directory method. I regularly run with an `MM_SERVICESETTINGS_SITEURL` defined, and could imagine others doing the same, but for whatever reason might still want to use the copying methodology.

That being said, copying is becoming less and less useful. I've already reorganized my meaningful GOPATH projects into directory hierarchies making these directories no longer peers. It might make sense to deprecate this altogether at some point.

lieut-data:
Thanks, @hanzei!

build/deploy/main.go
hanzei: It agree with all you said, but my comment was not about the code semantic. I was just commenting on the style.

My proposal is to rewrite it as:

```suggestion
		}
		
		if adminUsername != "" && adminPassword != "" {
```

lieut-data: Oh, I see! Makes sense :)

PR #89, opened 2020-03-18
Add junit.xml to gitignore

hanzei:
#### Summary

Don't add `junit.xml` to this repo

### DIFF
--- a/webapp/.gitignore
+++ b/webapp/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .npminstall
+junit.xml
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -103,8 +103,5 @@
       "<rootDir>/tests/setup.js"
     ],
     "testURL": "http://localhost:8065"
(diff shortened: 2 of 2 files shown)


lieut-data:
@hanzei, do you know what creates this file? The `jest-junit` configuration points at `build/test-results.xml`, so I'm wondering if something /else/ is misconfigured?

hanzei:
@lieut-data Good question. It get's generated by `jest-junit`. And you also also correct that https://github.com/mattermost/mattermost-plugin-starter-template/blob/6eea8c457727237c2e9b2c359cb6e1960c18c2a6/webapp/package.json#L107-L109 _should_ configure it in way that the output should be `build/test-results.xml`.

According to https://www.npmjs.com/package/jest-junit the correct parameter is `outputDirectory` not `output`.

Given that we put the test coverage report for `server` also into `server`, I'm proposing the same here. Hence, I just dropped the invalid setting.

PR #90, opened 2020-03-27
Integrate GolangCI-Lint

hanzei:
#### Summary
- Use plugin orb for CI. Given that `mattermost/plugin-ci` is the de facto standard for building plugins, I just replaced the generic build steps.
- Add golangci-lint config, update the Makefile to use it and fix any upcoming issues. I mostly copied the config from https://github.com/mattermost/mattermost-marketplace.


hanzei:
@agnivade @lieut-data Please take a look and think about this proposal. I have to rush to get this in soon.

hanzei:
cc @mattermost/integrations. Feel free to leave any feedback you have

.circleci/config.yml
```diff
@@ -1,27 +1,32 @@
...
+          filters:
+            tags:
+              only: /^v.*/
+      - plugin-ci/deploy-release-github:
+          filters:
+            tags:
+              only: /^v.*/
+            branches:
+              ignore: /.*/
+          context: matterbuild-github-token
```
lieut-data: This bit "worries" me -- the starter template is meant for endusers to take and use however they see fit, with a minimum of "bits" that have to be removed. Should we have a `mattermost-plugin-internal-starter-template` that uses the orb instead, and a "generic" one for the rest of the community?

.golangci.yml
```diff
@@ -0,0 +1,61 @@
...
+  modules-download-mode: readonly
+
+linters-settings:
+  goconst:
+    min-len: 2
+    min-occurrences: 2
+  gofmt:
+    simplify: true
+  goimports:
+    local-prefixes: github.com/mattermost/mattermost-starter-template
```
lieut-data: Will modifying this file be a pain when this repository is used as a template? Can we automate updating this?

.golangci.yml
```diff
@@ -0,0 +1,61 @@
...
+    - structcheck
+    - stylecheck
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+    - whitespace
+
+issues:
```
lieut-data: How does this all this behave if I remove the server component and distribute a webapp-only plugin? Is it correctly ignored?

server/plugin.go
```diff
@@ -21,7 +21,7 @@ type Plugin struct {
 }
 
 // ServeHTTP demonstrates a plugin that handles HTTP requests by greeting the world.
-func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Request) {
+func (p *Plugin) ServeHTTP(_ *plugin.Context, w http.ResponseWriter, _ *http.Request) {
```
lieut-data: Is this a GoLang-CI requirement? I think we should disable this rule, since this reduces the utility of the template if I have to fill-in the variable names. This also doesn't seem to be idiomatic -- the example on https://golang.org/pkg/net/http/#example_ListenAndServe doesn't use an `_` even though the variable is unused.

.circleci/config.yml
```diff
@@ -1,27 +1,32 @@
...
   version: 2
-  untagged-build:
+  ci:
     jobs:
-      - lint
-      - test
+      - plugin-ci/lint:
+          filters:
+            tags:
+              only: /^v.*/
```
agnivade: Are there instances of tags _not_ starting with v ?

.golangci.yml
hanzei: We could add a "What do I have to do after I forked this Repo" section to the `README`, which explains that this line has to be changed.

If you agree, I would add this as a separate PR

.golangci.yml
hanzei: Yes. I did remove `server` and `make golangci-lint` did run without issues.

server/plugin.go
hanzei: It's required by `unparam`. I personally like the style, but I see that it reduces the utility. Removing it.

hanzei:
@prapti I'm unsure if there is much to QA review here. `make golangci-lint` passes on CI. Let me know if you have any questions.

prapti:
@hanzei I can do a quick smoke test to see nothing is explicitly broken.

PR #93, opened 2020-04-24
auto npm install on sha256(package.json) changes

lieut-data:
#### Summary
Use the existing `.npminstall` file to track the hash of `package.json` from the last successful `npm install`. This will automatically install new dependencies when they change in addition to avoiding unnecessary rebuilds.

I was burned by this for the last time today.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ apply:
 
 ## Runs golangci-lint and eslint.
 .PHONY: check-style
-check-style: webapp/.npminstall golangci-lint
+check-style: npminstall golangci-lint
 	@echo Checking for style guide compliance
 
 ifneq ($(HAS_WEBAPP),)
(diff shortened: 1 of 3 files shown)


lieut-data:
@ali-farooq0 or @crspeller, in lieu of QA review on what are just tooling changes, can I ask you to give this a quick spin for yourself to verify the behaviour after completing your review?

Makefile
```diff
@@ -61,22 +61,22 @@ ifneq ($(HAS_SERVER),)
 endif
 
 ## Ensures NPM dependencies are installed without having to run this all the time.
-webapp/.npminstall:
+.PHONY: npminstall
+npminstall:
```
crspeller, line 65: Could we just change the target to `webapp/package.json` instead of writing custom functionality? Granted that would check the timestamp and not the hash, but that's probably good enough right?

lieut-data: This sounds interesting -- can you elaborate on how it would be implemented? Today, we use the existence of the `webapp/.npminstall` file to skip the `npm install` check, but it's a one way-switch: there's no timestamp awareness. 

Is there a feature in `make` to tie the build event to a cached timestamp?

crspeller: Sorry. I think I misunderstand the point of this PR. It was already setup to only run `npm install` when `package.json` is changed. What is this PR trying to accomplish?

crspeller: This is the alternative: https://github.com/mattermost/mattermost-plugin-starter-template/pull/94
Does it work for you?

lieut-data: Hah, yeah, that works! I suppose it's been too long since I've actually had to use make to intelligently manage dependency compilation :)

Also, that's an epic amount of compression relative to this pull request!

lieut-data:
Closing in favour of https://github.com/mattermost/mattermost-plugin-starter-template/pull/94.

PR #94, opened 2020-04-27
Fix makefile to npm install when package.json changes.

crspeller:
#### Summary
Fixes makefile so if package.json is modified then `npm install` will be re-run.

Alternative to: https://github.com/mattermost/mattermost-plugin-starter-template/pull/93


lieut-data:
Skipping `2: QA Review` on this since it's been developer tested (and impacts developer tooling only).

PR #97, opened 2020-05-21
Fix recursive copies in Makefile for OSX

grubbins:
#### Summary
Fixes incorrect location of public files when building.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,15 +103,15 @@ ifneq ($(wildcard $(ASSETS_DIR)/.),)
 	cp -r $(ASSETS_DIR) dist/$(PLUGIN_ID)/
 endif
 ifneq ($(HAS_PUBLIC),)
-	cp -r public/ dist/$(PLUGIN_ID)/
+	cp -r public dist/$(PLUGIN_ID)/
 endif
 ifneq ($(HAS_SERVER),)
-	mkdir -p dist/$(PLUGIN_ID)/server/dist;
-	cp -r server/dist/* dist/$(PLUGIN_ID)/server/dist/;
+	mkdir -p dist/$(PLUGIN_ID)/server
+	cp -r server/dist dist/$(PLUGIN_ID)/server/
 endif
 ifneq ($(HAS_WEBAPP),)
-	mkdir -p dist/$(PLUGIN_ID)/webapp/dist;
-	cp -r webapp/dist/* dist/$(PLUGIN_ID)/webapp/dist/;
+	mkdir -p dist/$(PLUGIN_ID)/webapp
+	cp -r webapp/dist dist/$(PLUGIN_ID)/webapp/
 endif
 	cd dist && tar -cvzf $(BUNDLE_NAME) $(PLUGIN_ID)
 


Makefile
```diff
@@ -92,7 +92,7 @@ ifneq ($(wildcard $(ASSETS_DIR)/.),)
 	cp -r $(ASSETS_DIR) dist/$(PLUGIN_ID)/
 endif
 ifneq ($(HAS_PUBLIC),)
-	cp -r public/ dist/$(PLUGIN_ID)/
+	cp -r public/ dist/$(PLUGIN_ID)/public/
```
hanzei, line 95: Could you elaborate a bit why this is needed? What is the bug this code change fixes?

grubbins: The bug is that files in `public/` don't actually get served by mattermost - they go into the tarball and are ignored by mattermost.

If we put them in the `public/` dir in the packaged plugin tarball, then mattermost starts hosting them under the plugin's URL, which is what we are meant to be demonstrating here.

hanzei: @grubbins On my local machine (Linux) the generated tarball is exactly the same with tar ball. Maybe this is a shell specific issue?

hanzei: https://github.com/mattermost/mattermost-plugin-starter-template/pull/97#discussion_r436647294

hanzei: On my local machine (linux, bash) both your branch and `master` generate exactly the same tarball. Maybe the issue you describe is caused by a shell specific behaviour?

grubbins: Aha, you're right, the issue occurred on OSX, which behaves differently to Linux - see `man cp` on OSX:
```
If the source_file ends in a /, the contents of the directory are copied rather than the directory itself.
```

Another version that I believe would work for both:
```
cp -R public dist/$(PLUGIN_ID)/
```
IMO this is actually clearer too. When I see the trailing slash in the first argument I have to think about the effect (e.g. `rsync` treats that differently even on Linux). With no trailing slash it's very clear that we are copying the dir itself to the destination.

(note `-R` does the same as `-r` on Linux, but on OSX, `-r` is deprecated in favour of `-R` - hence suggesting `-R`)

hanzei: Thanks for investigating. :+1: for omitting the trailing slash.

Could you please check if there are other instances of this issue in `Makefile`?

grubbins:
@hanzei I replaced all the recursive copies with the `cp -r <src>/* <dest>` style so they're consistent and they will work on OSX and Linux.

Makefile
```diff
@@ -100,18 +100,20 @@ bundle:
 	mkdir -p dist/$(PLUGIN_ID)
 	cp $(MANIFEST_FILE) dist/$(PLUGIN_ID)/
 ifneq ($(wildcard $(ASSETS_DIR)/.),)
-	cp -r $(ASSETS_DIR) dist/$(PLUGIN_ID)/
+	mkdir -p dist/$(PLUGIN_ID)/$(ASSETS_DIR)
+	cp -r $(ASSETS_DIR)/* dist/$(PLUGIN_ID)/$(ASSETS_DIR)/
```
hanzei, line 104: That change did broke CI as I doesn't work if`$ASSETS_DIR` is empty: https://app.circleci.com/pipelines/github/mattermost/mattermost-plugin-starter-template/340/workflows/87cc7120-9614-4702-9edf-b7b3607f05f3/jobs/717/parallel-runs/0/steps/0-103

grubbins: Mmm ok, yes I had thought the wildcard check above was already protecting against that case - evidently not.

grubbins: I suppose `ifneq ($(wildcard $(ASSETS_DIR)/.),)` actually says "if `ASSETS_DIR` is a directory".

grubbins: Then probably it's better to remove the `*` from all the recursive copies. I will test that locally (a bit more thoroughly)

grubbins:
OK, tested this version on both OSX and Linux with same result:

```
> tree -a dist/
dist/
├── com.mattermost.plugin-starter-template
│   ├── assets
│   │   └── .gitkeep
│   ├── plugin.json
│   ├── public
│   │   └── hello.html
│   ├── server
│   │   └── dist
│   │       ├── plugin-darwin-amd64
│   │       ├── plugin-linux-amd64
│   │       └── plugin-windows-amd64.exe
│   └── webapp
│       └── dist
│           └── main.js
└── com.mattermost.plugin-starter-template-0.1.0.tar.gz
```

I also checked behaviour if `public` or `$ASSETS_DIR` are empty or entirely missing - all seems to work as expected.

PR #98, opened 2020-05-21
Add .idea to gitignore

grubbins:
For the GoLand fans!

Also indicate directories for the items that are directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,10 @@
-dist
+dist/
 
 # Mac
 .DS_Store
 
 # VSCode
-.vscode
+.vscode/
+
+# Jetbrains
+.idea/


.gitignore
```diff
@@ -1,7 +1,10 @@
...
 
 # Mac
 .DS_Store
 
 # VSCode
-.vscode
+.vscode/
+
+# Jetbrains
+.idea/
```
hanzei, line 10: Well, I'm wondering if there should be an ignore rule for all editors. I'm leaning towards  recommending [a global gitignore file](https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) just because it's impossible to stay up to date with every possible configurations and editor. With that in mind `.vscode/` should also be removed. 

What do you think?

grubbins: Yeah that's fair, I would also be happy to have neither of them in there.

In general I like that approach - the ignore file should only deal with files that are expected to appear during normal dev/test work regardless of your IDE or OS.

marianunez: Agreed the global gitignore would be the cleaneast approach. We can remove `.vscode/` and possibly add it as a recommendation in the developer setup. 

However, I think it would be ideal if this could be done across Mattermost repos. Webapp it is currently including it in as well 
https://github.com/mattermost/mattermost-webapp/blob/f589dffb01621d89cbd5641f97d60a838dd11f81/.gitignore#L11

hanzei: Good point @marianunez. :+1: 

@grubbins Would you please remove `vscode` in this repo? Would you also be interested in removing it in the webapp?

grubbins: OK

hanzei:
Reopening per https://github.com/mattermost/mattermost-plugin-starter-template/pull/102#issuecomment-663916550

hanzei:
@marianunez Please take a look

PR #118, opened 2020-08-24
Drop interfacer and goconst from golangci-lint config

hanzei:
#### Summary
Both linter haven't been proven useful and created more false positives then correct results.

#### Ticket Link
https://community.mattermost.com/core/pl/r5fdx8xmp3na8er87zkt5pyqpy

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,7 +33,6 @@ linters:
     - gosimple
     - govet
     - ineffassign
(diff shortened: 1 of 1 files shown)


agarciamontoro (approved):
I have not struggled with `goconst` yet, but I did with `interfacer`. Thank you for this!

Issue #115, opened 2020-08-21
Trailing slash on `MM_SERVICESETTINGS_SITEURL` causes non-obvious errors


shieldsjared:
While attempting to use this starter template, I set my `MM_SERVICESETTINGS_SITEURL` with a trailing slash.  It was not immediately obvious that this was the cause of the error I was receiving, as it _appeared_ that some calls were successful while others were not

```
2020/08/21 13:00:02 Authenticating using token against https://mattermost/.
2020/08/21 13:00:02 Uploading plugin via API.
2020/08/21 13:00:02 Enabling plugin.
Failed: failed to enable plugin: : Sorry, we could not find the page., There doesn't appear to be an api call for the url='/api/v4/plugins/resfor-mattermost-azure-devops-integration/enable'.  Typo? are you missing a team_id or user_id as part of the url?

Usage:
    pluginctl deploy <plugin id> <bundle path>
    pluginctl disable <plugin id>
    pluginctl enable <plugin id>
    pluginctl reset <plugin id>
make: *** [Makefile:127: deploy] Error 1
```

As you can see in the log, the upload _appears_ to work, but then the subsequent call to enable it fails. What's interesting is that I dug into the logs and found that the upload actually did not succeed at all... This led to quite a bit of chin scratching and hunting through code before realizing I should remove the trailing slash.

Not sure if the best solution is to
1. Trim the trailing slash for the end-user, or
2. Put a comment in the readme file (easiest)

Happy to submit a PR for either.

hanzei:
Trimming the trailing slash is a fine solution. Looking forward to your PR! :+1:

shieldsjared:
@hanzei - PR Submitted.

PR #120, opened 2020-08-31
Removed trailing slash from site url.

shieldsjared:
#### Summary
If the `MM_SERVICESETTINGS_SITEURL` environment variable is set with a trailing slash, the API calls made to upload the plugin result in unpredictable results.  Code added simply removes any trailing slash included in the env variable.

Probably could have migrated all of the siteURL related code to get/validate/trim the siteURL into a function, but not sure it was really necessary for the overall scope of `pluginctl`.

#### Ticket Link
Fixes issue described in #115.


hanzei (approved):
LGTM :+1:

build/pluginctl/main.go
```diff
@@ -80,6 +81,8 @@ func getClient() (*model.Client4, error) {
 		return nil, errors.New("MM_SERVICESETTINGS_SITEURL is not set")
 	}
 
+	siteURL = strings.TrimSuffix(siteURL, "/")
```
lieut-data, line 84: This is fine for `pluginctl`, but as you hinted in the summary, sounds like its a "bigger problem" that might be better addressed in `model.NewAPIv4Client`. 

Can you confirm that all Go client usage of the API is broken when there's a trailing slash? If so, would it make sense to strip that trailing slash in `model.NewAPIv4Client` in the server code instead?

shieldsjared: Great point... I actually started digging deeper to figure out specifically why the call to `client.UploadPluginForced` claims it succeeds but really doesn't... 

I'm still digging, but so far, here is what I see:
1. When there is a trailing slash, that call generates  POST request, but the server is logging a GET and returning a 200 (so the `deploy()` routine believes it was successful and moves on to the next call to enable the plugin (which fails). I find an entry like this in my server logs:

```
{"level":"debug","ts":1599004482.4648635,"caller":"web/handlers.go:100","msg":"Received HTTP request","method":"GET","url":"/api/v4/plugins","request_id":"936f59oqg7d57y9hr94hhnhizc","status_code":"200"}
```

- When there is NOT a trailing slash, the call generates a POST request, and the server correctly logs a POST and returns a 201... everything proceeds as it should from here... I find an entry like this in my server logs:

```
{"level":"debug","ts":1599004263.3228014,"caller":"web/handlers.go:100","msg":"Received HTTP request","method":"POST","url":"/api/v4/plugins","request_id":"75rneq55d7ywuboo9b3ia5nrqa","status_code":"201"}
```

It's extremely odd that, when including the trailing slash, the server seems to be mis-handling it (and even logging it incorrectly too!).  At this point, I'm not 100% sold on where the problem lies, but wanted to get an update (haven't had too much time to put into it).  My guess is that there's quite a few things rearing their heads here... 

To your original question though - does all go client usage break... Yes. Calls, for example, to enable and disable the plugin will fail if there is a trailing slash (I'm running MM on a windows server, but that shouldn't matter).  So yes, if you're looking for the "premium" fix, I agree that we should put that fix over in `model.NewAPIv4Client`... I noticed that this project is still referencing an interesting commit w/in the mattermost-server repo... Can you point me into a direction as to where I should even submit that fix against?  I suppose we can close this PR here and open up an issue over there... Thoughts?

lieut-data: Thanks for the deep investigation, @shieldsjared! I suspect the behaviour you observed is a result of the server issuing a redirect from `//...` to `/...` without directing the client to resubmit the `POST` against the new endpoint.

My proposal to fix this:
* Open a PR in https://github.com/mattermost/mattermost-server modifying `model.NewAPIv4Client` with effectively the `siteURL = strings.TrimSuffix(siteURL, "/")` we already have. A quick unit test can verify these changes, and then all of our existing unit tests will exercise these APIs programmatically to verify nothing else is amiss.
* Keep this PR open, and once merged, just update the go module dependency to point at the new `master` commit. The `go.mod` looks weird because it's referencing a pseudo-version from the `master` branch vs. a tagged release -- namely commit `ed34468996e6` (from around July 23rd).

Does that make sense? Happy to help however I can :)

jasonblais: Hey @shieldsjared - let us know if you have any questions about Jesse's feedback above. We'd be happy to clarify :)

shieldsjared: Thanks @jasonblais It all makes sense :) Just haven't had a chance to pop over and execute next steps.  I'll get it done shortly, don't want to leave such a simple thing like this lingering. Sorry!

jasonblais: @shieldsjared All good! No worries, thanks for the quick reply :)

shieldsjared:
**phew** finally got a moment to submit that simple PR:

https://github.com/mattermost/mattermost-server/pull/15513

lieut-data:
Awesome, thanks @shieldsjared :)

jasonblais:
@shieldsjared Quick update (which you may have already seen): https://github.com/mattermost/mattermost-server/pull/15513 is merged so the remaining step is to 

> update the go module dependency to point at the new `master` commit. The `go.mod` looks weird because it's referencing a pseudo-version from the `master` branch vs. a tagged release -- namely commit `ed34468996e6` (from around July 23rd).

No rush by the way! I'll remove the `Lifecycle/1:stale` label since I know this was recently been worked on (via the mattermost-server PR).

build/pluginctl/main.go
shieldsjared: This code is OBE now. I went ahead and removed it from the PR.

shieldsjared:
@jasonblais - Updated the PR to remove the internal fix for the trailing slash, and instead pointing to the specific commit `e726b0426826` in `mattermost-server` that contains the fix w/in the client.  Ready for a fresh review :)

jasonblais:
@hanzei @lieut-data ready for a fresh review

hanzei (approved):
:rocket:

lieut-data (approved):

lieut-data:
Skipping QA review since already tested in the upstream server change.

lieut-data:
Thanks so much, @shieldsjared!

PR #127, opened 2020-09-18
Don't generate manifest.ts

hanzei:
#### Summary
Instead of generating `manifest.ts` via `make apply` the code simply reads it from the json file.

We might be able to drop `make apply` completely in the future.

#### Ticket Link
None


agarciamontoro (approved):
Neat! :tada:

jfrerich (approved):
Removing the need for `make apply` would be awesome!

The command is already tied in to the Make commands, but accessing plugin.json as a single source of truth will also be great 🎉

PR #129, opened 2020-09-23
Migrate index.js to use Typescript

hanzei:
#### Summary
Most of the code is stolen from https://github.com/mattermost/mattermost-plugin-msteams-meetings/. All credit goes to @larkox.

#### Ticket Link
None


larkox (approved):

webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface PluginRegistry {
```
hanzei, line 1: I would love to see us export such interface, which plugins then import.

jespino (approved):
LGTM

ali-farooq0 (approved):
👍

Issue #112, opened 2020-08-13
Build fails when removing the webapp component


Salatfreak:
According to the [README](/mattermost/mattermost-plugin-starter-template/blob/master/README.md#how-do-i-make-a-server-only-or-web-app-only-plugin) we can simply delete the `webapp` folders and remove the corresponding sections from `plugin.json`. `make` will produce the following error message though:

> make: *** No rule to make target 'webapp/package.json', needed by 'webapp/node_modules'.  Stop.

This could for example be fixed by always fulfiling the `webapp/package.json` target if the plugin doesn't have a webapp like so:
```make
## Always fulfil webapp/package.json if plugin doesn't have webapp
ifeq ($(HAS_WEBAPP),)
.PHONY: webapp/package.json
webapp/package.json: ;
endif
```

hanzei:
Good catch @Salatfreak. Are you intested in submitting a patch for this bug?

PR #130, opened 2020-09-23
Skip the `webapp/package.json` if it is missing

amwolff:
#### Summary

I've used the [wildcard function](https://www.gnu.org/software/make/manual/html_node/Wildcard-Function.html) to skip the `webapp/package.json` target dependency if it is missing.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-starter-template/issues/112.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ endif
 endif
 
 ## Ensures NPM dependencies are installed without having to run this all the time.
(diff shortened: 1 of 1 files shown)


hanzei (approved):
Tested locally and works nice fix :+1:

cpoile (approved):
Thanks @amwolff !

PR #149, opened 2021-04-19
Disable fieldalignment golangci-lint check

hanzei:
#### Summary
Disable the newly added `fieldalignment` check as it proves little value for new devs.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-starter-template/issues/148

PR #150, opened 2021-04-29
golangci: disable fieldalignment

lieut-data:
#### Summary
Disable `fieldalignment` for golangci -- this is an over optimization in most every use case and doesn't need to be on by default.

#### Ticket Link
None.


hanzei:
Duplicate of https://github.com/mattermost/mattermost-plugin-starter-template/pull/149?

lieut-data:
Thanks @hanzei!

PR #156, opened 2021-07-01
Add confirmation message when building with plugin id "com.mattermost.plugin-starter-template"

haardikdharma10:
#### Summary

Fixes: #152

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,7 @@ endif
 .PHONY: server
 server:
 ifneq ($(HAS_SERVER),)
+	$(info Are you sure you want to build the plugin with an id of "com.mattermost.plugin-starter-template"? Consider editing plugin.json to configure your project.)
 	mkdir -p server/dist;
 ifeq ($(MM_DEBUG),)
 	cd server && env GOOS=linux GOARCH=amd64 $(GO) build $(GO_BUILD_FLAGS) -o dist/plugin-linux-amd64;


Makefile
```diff
@@ -57,6 +57,7 @@ endif
 .PHONY: server
 server:
 ifneq ($(HAS_SERVER),)
+	$(info Are you sure you want to build the plugin with an id of "com.mattermost.plugin-starter-template"? Consider editing plugin.json to configure your project.)
```
haardikdharma10, line 60: Wondering if using `$(info message)` is the right approach or should it be changed to `$(warning message)`?

hanzei: Only my system using `warning` outputs the line number, which is not so helpful.

Maybe adding a newline after the output helps to separate it from the following log lines.

hanzei (approved):

hanzei:
Sorry for the delay @haardikdharma10

larkox (approved):

Makefile
larkox, line 60: This will show the message always, right? I mean... even if I have edited my plugin.json and have the correct id, I will still receive this message. Shouldn't we do a check on the manifest id somehow? Or add something more like "edit the plugin.json and remove this line".

larkox (commented):
One question here.

Makefile
haardikdharma10: Hey @larkox, thanks for your review:) Instead of having a check, we can do something simple like adding **"Ignore if already edited"** to the end of the message. How does that sound? 
/cc @hanzei

larkox: I don't like it, because a warning that always appear is a warning you ignore. And the problem will still be there.

haardikdharma10: What change would you like to suggest here?

larkox: Something like:
```
ifeq ($(PLUGIN_ID), com.mattermost.plugin-starter-template)
    echo "Are you sure you want to build the plugin with an id of "com.mattermost.plugin-starter-template"? Consider editing plugin.json to configure your project."
endif
```

I would even consider exiting and not allowing the plugin to be compiled, but that may be too drastic :P

Issue #179, opened 2023-04-06
GitHub workflow fails for fork


svenssonaxel:
After forking the starter template and pushing my plugin code to GitHub, the workflow keeps failing due to missing AWS keys. If I need or should want such keys, then why, otherwise what's the recommended way to deal with it? Either way, could we add a section to `README.md` about it?

hanzei:
@spirosoik We briefly talked about this issue when you did the migration. We need to ensure that the Workflow is usually for community members without any further changes. Can you please take a look what causes the workflow to fail on a fork?

spirosoik:
@toninis could you have a look here please?

toninis:
@svenssonaxel can you please share the repo you are facing the issue ? Maybe we 've missed the fix there.

kaakaa:
I encountered same issue in https://github.com/kaakaa/mmplugin-ci-test/actions that is created from this template.
In [Actions](https://github.com/kaakaa/mmplugin-ci-test/actions) tab, ci [#1](https://github.com/kaakaa/mmplugin-ci-test/actions/runs/5057742931) and [#2](https://github.com/kaakaa/mmplugin-ci-test/actions/runs/5057743037) failed, and [#3](https://github.com/kaakaa/mmplugin-ci-test/actions/runs/5057802643) is the result after merging https://github.com/mattermost/mattermost-plugin-starter-template/pull/184/commits/50562ede056b1cc892c0a8015af50734e446adad but it also failed. So #184 doen't seem to fix this issue.

Just idea: [`plugin-ci/delivery`](https://github.com/mattermost/actions-workflows/blob/main/.github/workflows/plugin-ci.yml#L78) job needs `secrets.PLUGIN_AWS_ACCESS_KEY_ID` and `secrets.PLUGIN_AWS_ACCESS_KEY_ID` to run, so it would be possible to fix this by checking if those secrets was set before running `plugin-ci/delivery` ([here](https://github.com/mattermost/actions-workflows/blob/main/.github/workflows/plugin-ci.yml#L79)). (Even if those secrets was set, `plugin-ci/delivery` from a community member will fail because they don't have permissions to write archives into `mattermost-plugins-ci/ci/` bucket.)

[...]

PR #184, opened 2023-05-22
fix: Fix conditions for using secrets on forks.

toninis:
#### Summary

This should fix any future plugin creations from this template.

Fixes https://github.com/mattermost/mattermost-plugin-starter-template/issues/179

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,13 +1,16 @@
 name: cd
 on:
   workflow_run:
-    workflows: ["ci"]
-    branches-ignore: ["*"]
+    workflows:
+      - ci
+    branches:
+      - master
+      - main
     types:
       - completed
   push:
     tags:
-      - "v*"
+      - "v[0-9]+.[0-9]+.[0-9]+"
 
 permissions:
   contents: read
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,8 +5,9 @@ on:
   push:
     branches:
       - master
+      - main
     tags:
-      - "v*"
+      - "v[0-9]+.[0-9]+.[0-9]+"
   pull_request:
 
 permissions:


.github/workflows/cd.yml
```diff
@@ -1,13 +1,16 @@
 name: cd
 on:
   workflow_run:
-    workflows: ["ci"]
-    branches-ignore: ["*"]
+    workflows:
+      - ci
+    branches:
+      - master
+      - main
```
hanzei, line 8: Does that mean the `cd` workflow only runs on these branches? Or does it also run on tags?

hanzei (changes requested):
Thanks for taking a look at the contributor's issue :+1: 

Can you please elaborate on how this PR fixes the original issue of missing AWS keys?

.github/workflows/cd.yml
toninis: It will run on tags also since it has the directive below
```yaml
  push:
    tags:
      - "v[0-9]+.[0-9]+.[0-9]+"
```

toninis:
> Thanks for taking a look at the contributor's issue 👍
> 
> Can you please elaborate on how this PR fixes the original issue of missing AWS keys?

AWS keys are not missing they just cannot be accessed from forked PRs . This ensures that access to AWS keys is only allowed when CI is running on master/main or tags . We will not run the CD part during the PR so you won't have this issue.

hanzei:
> This ensures that access to AWS keys is only allowed when CI is running on master/main or tags .

Does that mean CD runs on `master/main`? Does that cause an issue if the AWS keys are not yet?

On another note: The previous circleCI pipeline that was configured in this repos allowed community members to simply fork the repo and start using it -  even for building their own releases- E.g. https://github.com/matterpoll/matterpoll/blob/master/.circleci/config.yml is one of the community projects. How can they use the new GH actions workflow to cut releases without requiring an AWS account?

toninis:
> 

Still causes this issue for community repos outside Mattermost org . I do not know why we enforce AWS push on the plugins since I was not part of the decision. In order to facilitate also community we should remove the AWS part completely since this is something we use . 

We can just comment out the `cd` part and conditionally add it if the plugin is maintained from within Mattermost org . 0/5 on this since I do not have any requirements on this project.

.github/workflows/cd.yml
```diff
@@ -1,13 +1,16 @@
...
     types:
       - completed
   push:
     tags:
-      - "v*"
+      - "v[0-9]+.[0-9]+.[0-9]+"
```
hanzei, line 13: Some plugins use build tags like in https://github.com/mattermost/mattermost-plugin-github/releases/tag/v2.1.6-rc1. Will `cd` still trigger for builds like that?

toninis: No we also need to include the rc ones

toninis:
Colsing this for now . If we have any other issue we will address it .

PR #192, opened 2023-12-12
Sync with playbooks: install-go-tools, gotestsum, and dynamic versions

lieut-data:
#### Summary
This brings into the starter-template a handful of changes originally introduced with Playbooks:
* an automatic `install-go-tools` target that manages a local, versioned copy of any `go install`ed tools needed
* adoption of `gotestsum` for improved go testing semantics
* reverting the embedding of `plugin.json`, restoring `make apply`, and shifting to a dynamically generated version number based on the nearest tag.

This last point is easier explained by linking to the original PR: https://github.com/mattermost/mattermost-plugin-playbooks/pull/433, but I note that the reason I reverted embedding `plugin.json` is that this root file is no longer the final source of truth: rather, it's a (big) input into the final manifest that includes the dynamically generated version. Open to different approaches, but either way we don't have to care about checking in the generated manifests: they remain generated and `.gitignore`d.

In this repository, since we have yet to tag a version, any build will result in `v0.0.0+<commit hash>`. Any build on a tagged version will be `vX.Y.Z` exactly. And any downstream commit from there would be `vX.Y.Z+<short hash>` until we tag a new version.

#### Ticket Link
None.


cpoile (approved):
I think this is a good way to do it, nice!

build/manifest/main.go
```diff
@@ -62,6 +111,30 @@ func findManifest() (*model.Manifest, error) {
...
+			version = "v0.0.0+" + BuildHashShort
+		}
+	}
+	manifest.Version = strings.TrimPrefix(version, "v")
+
+	// Update the release notes url to point at the latest tag, if present.
+	if BuildTagLatest != "" {
+		manifest.ReleaseNotesURL = manifest.HomepageURL + "releases/tag/" + BuildTagLatest
+	}
+
```
mickmister, line 137: LGTM :+1:

If I understand correctly:
- if the current commit matches a tag, then we use that
- If not, use the latest tag and append the latest commit hash
- If no tag available, use `v0.0.0` and append the latest commit hash

mickmister (approved):

build/manifest/main.go
lieut-data: 💯

build/setup.mk
```diff
@@ -4,8 +4,13 @@ ifeq ($(GO),)
     $(error "go is not available: see https://golang.org/doc/install")
 endif
 
+# Gather build variables to inject into the manifest tool
+BUILD_HASH_SHORT = $(shell git rev-parse --short HEAD)
+BUILD_TAG_LATEST = $(shell git describe --tags --match 'v*' --abbrev=0)
```
hanzei, line 9: Any way we can avoid logging the error if no tag was found? (`No names found, cannot describe anything.`)

build/manifest/main.go
```diff
@@ -62,6 +111,30 @@ func findManifest() (*model.Manifest, error) {
...
+		}
+	}
+	if version == "" {
+		if BuildTagLatest != "" {
+			version = BuildTagLatest + "+" + BuildHashShort
+		} else {
+			version = "v0.0.0+" + BuildHashShort
+		}
+	}
+	manifest.Version = strings.TrimPrefix(version, "v")
```
hanzei, line 131: I quite like the version bump PRs as they provide a rendezvous for all parties involved in the release process. That rendezvous point is not so important for plugins maintained by MM, but for Community-maintained ones, where communication happens mostly on GitHub, it is helpful.

With that in mind, I wonder if there is a way we can allow plugin devs to pick if they want dynamic versions or file defined ones, e.g. only use dynamic versions if the `version` field is empty in the file. This way, it's an opt-in feature, which preserves the workflow for Community-maintained plugins.

hanzei (changes requested):
Thanks for improving the starter template :rocket:

build/setup.mk
lieut-data: Indeed, fixed!

Makefile
```diff
@@ -188,17 +200,28 @@ detach: setup-attach
...
+	$(GOBIN)/gotestsum -- -v ./...
+endif
+ifneq ($(HAS_WEBAPP),)
+	cd webapp && $(NPM) run test;
+endif
+
+## Runs any lints and unit tests defined for the server and webapp, if they exist, optimized
+## for a CI environment.
+.PHONY: test-ci
+test-ci: apply webapp/node_modules install-go-tools
```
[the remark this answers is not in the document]
lieut-data: Not at the moment, but I'm a bit blurry on how to achieve that -- it looks like we import `@main`, which would make whatever changes we add there affect all plugins at the same time, vs. plugins opting in to updated CI incrementally. Not sure how to crack this nut: thoughts?

Makefile
```diff
@@ -41,24 +43,34 @@ endif
...
 
+## Propagates plugin manifest information into the server/ and webapp/ folders.
+.PHONY: apply
+apply:
+	./build/bin/manifest apply
+
+## Install go tools
+install-go-tools:
+	@echo Installing go tools
+	$(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.1
```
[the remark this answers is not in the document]
lieut-data: See below: uncertain how to roll out these changes to CI without breaking everything. For now, this at least gives the developer a reliable definition of the version in play. Thoughts?

build/manifest/main.go
```diff
@@ -62,6 +111,32 @@ func findManifest() (*model.Manifest, error) {
 		return nil, errors.Wrap(err, "failed to parse manifest")
 	}
 
+	// If no version is listed in the manifest, generate one based on the state of the current
+	// commit, and use the first version we find (to prevent causing errors)
+	if manifest.Version == "" {
```
hanzei, line 116: Should the documentation make it clear that plugin devs can choose to use git-based versions or file-based versions?

Makefile
hanzei, line 54: Do we actually need to install the binary? Or is 
```suggestion
	$(GO) run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.1
```
enough?

Makefile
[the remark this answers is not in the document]
hanzei: Maybe it's time to move to hand-picked CI versions. We have been bitten by the auto-update of @main way too often. Espcecially, updates to `golangci-lint` caused issues in the past.

cc @mickmister

hanzei (approved):
A few open discussion points, but nothing blocking

Makefile
lieut-data: Oh, that would be convenient! But unfortunately much slower it seems?

CleanShot 2024-01-23 at 09 47 36@2x

build/manifest/main.go
lieut-data: Good idea! Updated README.md.

Makefile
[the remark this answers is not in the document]
hanzei: @mickmister Would you be open to owning that change? Brightscout could surely help with executing it.

Makefile
hanzei: Well, the added duration comes from checking the remote go modules DB. The increase is similar running `go install`.

One downside of `go run` is, that it errors if the machine is offline.

lieut-data: Ah, you're right: I wasn't comparing apples to apples as I thought!

PR #196, opened 2024-02-11
fix: replace any backticks (`) in manifestStr

NeilHanlon:
#### Summary
if manifestStr (read from plugin.json) has any '`' characters in it, such as when using markdown formatting in a text value, the generated go code fails to build due to unescaped '`' characters, leading the compiler to believe some input is mismatched.

This change passes the manifestStr through a Replacement filter to replace the ` characters with input which will be properly formatted through the compiler.

#### Ticket Link
n/a

### DIFF
--- a/build/manifest/main.go
+++ b/build/manifest/main.go
@@ -160,6 +160,9 @@ func applyManifest(manifest *model.Manifest) error {
 		}
 		manifestStr := string(manifestBytes)
 
+		// replace any backtiks (`) in manifestStr with suitable replacement for multiline literal
+		manifestStr = strings.ReplaceAll(manifestStr, "`", "` + \"`\" + `")
+
 		// write generated code to file by using Go file template.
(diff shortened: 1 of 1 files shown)


hanzei (changes requested):
Thanks for the contribution @NeilHanlon :+1: 

While the fix seems reasonable, I would like to see a test that verifies that the newly generated manifest is valid.

NeilHanlon:
> Thanks for the contribution @NeilHanlon 👍
> 
> While the fix seems reasonable, I would like to see a test that verifies that the newly generated manifest is valid.

Absolutely! I will add some tests later--I also found an issue with the typescript generator, as well.. Will have to think about that a bit, unless you have any immediate thoughts on how to support markdown in plugin.json (e.g. https://github.com/rocky-linux/mattermost-plugin-community-toolkit/blob/master/plugin.json)

PR #207, opened 2024-07-17
Sim Assignment: MM-NVD

Illbjorn:
# Summary

This PR contains a very basic implementation of a Mattermost plugin (`mm-nvd`) which can be configured via [slash commands](#basic-usage) to monitor the [NIST Vulnerabilit Database (NVD) Feed API](https://nvd.nist.gov/developers/vulnerabilities) and announce newly released vulnerabilities as they're discovered.

Example posts produced by this plugin (the `subscribe` response and an actual CVE announcement):

# Basic Usage

| Command | Purpose | Notes |
|---------|---------|-------|
| `/nvdm subscribe` | Subscribes a worker to the channel the command was invoked in. The worker will then produce notifications of any _new_ CVEs to this channel as they arrive. |
| `/nvdm unsubscribe` | Unsubscribes a worker from the channel the command was invoked in. |
| `/nvdm set key=value` | Sets a configurable value associated with the NVD query. This can refine results produced to only those including specific keywords, above certain severities, etc. |
| `/nvdm unset key` | Unsets a configurable value associated with the NVD query. <br>**NOTE: This is not yet implemented, see the various `TODO`s.** |

Currently configurable settings via `/nvdm set` include:

| Key | Expected Value | Purpose |
|-----|----------------|---------|
| `cvssv2severity` | `LOW`, `MED` or `HIGH` | Adjusts the NVD query for the current channel to only produce results at or above the provided severity. |
| `keywordsearch` | Any | Adjusts the NVD query for the current channel to only produce results including the provided keywords. |
| `cvetag` | `disputed`, `unsupported-when-assigned` or `exclusively-hosted-service` | Adjusts the NVD query for the current channel to only produce results including the provided tag. |

# Package Overview

This plugin is comprised of three major packages:
1. `nvd` @ `nvd`
2. `nvdm` @ `server/internal/nvdm`
3. `nvdp` @ `server/internal/nvdp`

## Package: NVD

NVD defines the simple query builder and wrapper for interacting with the NVD API.

## Package: NVDP

NVDP ("NVD Plugin") holds the Mattermost plugin itself.

## Package: NVDM

NVDM ("NVD Monitor") holds the CVEWatcher and CVEWatcherGroup which serve as the worker(s) and manager for configured NVD queries. Package `NVDP` will register workers via the WatcherGroup's `Register()` method using the Mattermost channel ID as the "unique ID".

# Notes

There are some major pieces left unfinished:
- Command autocomplete is not implemented - there's a lot in the way of UX that just isn't where I'd like it to be.
- The `/nvdm unset` command is not implemented.
- A large swath of the `nvd` package's capability remains unimplemented by packages `nvdm` and `nvdp`.
- **Unit tests need to be implemented!**
- Plugin configuration and data persistence are not implemented.
- Much more - there are plenty of `TODO`s in the code comments.


nvd/api.go
fmartingr, line 1: This package would easily be it's own library. I'm not sure if this was created for the sim, but if that was the case you spent a significant amount of time here.

server/internal/nvdm/watcher.go
```diff
@@ -0,0 +1,61 @@
...
+	cveQuery *nvd.CVEQuery
+	// The last time this worker polled the NVD.
+	lastCheck time.Time
+}
+
+func (w *CVEWatcher) CVEQuery() *nvd.CVEQuery {
+	return w.cveQuery
+}
+
+func (w *CVEWatcher) Run() {
```
fmartingr, line 33: FYI: https://developers.mattermost.com/integrate/plugins/components/server/ha/

server/internal/nvdp/watchers.go
```diff
@@ -0,0 +1,101 @@
...
+
+// WatcherUnsubscribe unsubscribes (discontinues CVE announcements) for the
+// channel the command was invoked in.
+func (p *Plugin) WatcherUnsubscribe(args *model.CommandArgs) {
+	// only attempt unsubscribe if we've actually registered a worker for this
+	// channel
+	if worker := p.cveWG.Worker(args.ChannelId); worker != nil {
+		// perform the unregister
+		if err := p.cveWG.Unregister(args.ChannelId); err == nil {
+			// TODO: really should be doing something with these errors.
```
fmartingr, line 44: What would you do?

server/internal/nvdp/watchers.go
```diff
@@ -0,0 +1,101 @@
...
+			})
+		return
+	}
+
+	// get the worker's query
+	query := worker.CVEQuery()
+
+	// iterate our params map by key-value pair
+	for k, v := range params {
+		// TODO: there's absolutely a better way to do this. Time constraints!
```
fmartingr, line 85: Care to explain? 🤓

fmartingr (commented):
I left some comments here and there. Didn't focus too much on the code style and other less relevant things due to the time constraints, but left some questions around for you.

nvd/api.go
AX-AMaxwell: This can absolutely be its own library. But it was created within the time constraints specified of the sim (I think I went right up to 5hrs.).

Illbjorn: It could absolutely be it's own library - this is why I implemented it at the root of the directory!

This was actually implemented within the time constraints (right about 5hrs.). I always make the joke the best way you can improve as a dev is to learn to type faster! Haha.

server/internal/nvdm/watcher_group.go
```diff
@@ -0,0 +1,92 @@
...
+		ctx:          cctx,
+		cancel:       cancel,
+		pollInterval: &wg.pollingRate,
+		handler:      cb,
+		cveQuery:     query,
+		lastCheck:    time.Now(),
+	}
+
+	// begin the worker
+	go w.Run()
```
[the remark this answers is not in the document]
Illbjorn: Yes, all watchers are initially registered with an unfiltered query. The idea of the "setup" process would be:
```
/nvd subscribe
/nvd set cvssv2severity=HIGH # now we have a filtered query applied
```

The plugin is also designed to only allow a single `subscribe` call (a single watcher/query) per-channel. So I suppose the envisioned implementation of this by a customer would be, as an exmaple, perhaps three Mattermost channels: 'CVEs-Critical', 'CVEs-Microsoft', 'CVEs-Android' and in each of these the `/nvd set` criteria would be customized to provide a unique, specific feed of information.

server/internal/nvdp/watchers.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/illbjorn/mm-nvd/nvd"
+)
+
+// WatcherSubscribe initiates the process of announcing CVEs to the channel the
+// command was invoked in.
+func (p *Plugin) WatcherSubscribe(args *model.CommandArgs) {
+	// don't register the same channel twice
+	if p.cveWG.Worker(args.ChannelId) == nil {
+		// register a callback to fire a message to this channel when matching
```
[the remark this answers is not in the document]
Illbjorn: "Matching criteria" here means the `CVEQuery` filter criteria. So that would be whatever the consumer has set with `/nvd set tag=key`.

server/internal/nvdp/watchers.go
Illbjorn: I left these comments more so to indicate that I know something _should_ be done with them but I didn't understand well enough how Mattermost's error handling and logging works (ex: is there a package in the Mattermost Go packages that can log to the System Console?) to implement something here.

Just in general: I'd be logging them, shipping them to a log aggregation stack, things like these.

server/internal/nvdm/watcher_group.go
```diff
@@ -0,0 +1,92 @@
...
+// CVEWatcherGroup.Register() to be called when a new CVE matching that worker's
+// nvd.CVEQuery has been identified.
+type NewCVEHandler func(cve nvd.CVE)
+
+// NewCVEWatcherGroup initializes a CVEWatcherGroup with some default values.
+func NewCVEWatcherGroup(ctx context.Context) *CVEWatcherGroup {
+	return &CVEWatcherGroup{
+		ctx:         ctx,
+		pollingRate: 10 * time.Second,
+		workers:     make(map[string]*CVEWatcher, 64),
```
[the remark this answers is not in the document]
Illbjorn: >  How would you improve this? What would happen if this plugin is run on a Mattermost server that has High Availability enabled? (That is, multiple Mattermost processes/pods)

I need to sit down and read through your HA doc above but my initial response would be to implement a sync.Mutex and lock/unlock where appropriate.

>  What would happen if you try to setup the 65th watcher?

Originally I would've thought a panic. But doing some testing it looks like it'll just continue to dynamically resize the underlying array until you run out of memory.

hanzei (changes requested):
Hi @Illbjorn,

This is an quite interesting plugin, it seems quite useful to me :+1: 

Would you be open to submitting your plugin to the Mattermost Marketplace via [this form](https://docs.google.com/forms/d/e/1FAIpQLSfWeNZgLlciYFXrAQ9CvH-BxcT54Y5b_9uNnLxs7E1K5tnawQ/viewform)?

This repo should stay as a template. Your plugin code could stay in your own repo.

PR #208, opened 2024-09-19
feat(e2e): playwright base setup

saturninoabril:
#### Summary

This is mainly based from the awesome works by @rahulsuresh-git and @mickmister at https://github.com/mattermost-community/mattermost-plugin-todo/pull/228. Thanks to both of you!

This plugin template is used to get started in developing plugins for Mattermost, and to make it easier for the developers to write E2E tests, this PR introduces the base setup for Playwright with reference from the `mattermost` repo ([here](https://github.com/mattermost/mattermost/tree/master/e2e-tests/playwright)), taking advantage of the patterns/models/setup readily available from there.

Screenshot 2024-09-19 at 6 46 58 PM


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: ci/install-plugin-webapp-deps
+        run: cd webapp && npm ci
+
+      - name: ci/build-and-install-plugin
+        run: make deploy
+
+      - name: ci/checkout-mattermost
+        run: |
+          git clone https://github.com/mattermost/mattermost.git ../mattermost
```
hanzei, line 97: Should we use `--depth ` here as well?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,149 @@
...
+          npm run check
+
+      - name: ci/run-e2e-tests
+        run: |
+          cd e2e
+          npm run test-ci
+        env:
+          CI: true
+          NODE_OPTIONS: '--no-experimental-fetch'
+          PW_ENSURE_PLUGINS_INSTALLED: com.mattermost.plugin-starter-template
```
hanzei, line 136: Can we extract the plugin ID programmatically?

e2e/.nvmrc
```diff
@@ -0,0 +1 @@
+20.11
```
hanzei, line 1: The webapp code of the plugin still used node 16. (https://github.com/mattermost/mattermost-plugin-starter-template/blob/master/.nvmrc#L1)

I'm hesitant that the use of different node versions will cause pain for developers. Can we align the two versions?

e2e/.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "bracketSpacing": false,
+    "printWidth": 120,
+    "singleQuote": true,
+    "tabWidth": 4
+}
```
hanzei, line 6: 0/5 should we use a global confirmation for all webapp node and not only the e2e code?

e2e/support/constant.ts
```diff
@@ -0,0 +1 @@
+export const starterTemplatePluginId = 'com.mattermost.plugin-starter-template';
```
hanzei, line 1: Same question here. Can we extract the plugin ID?

e2e/tests/basic.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    // # Visit system console
+    const systemConsolePage = new pages.SystemConsolePage(page);
+    await systemConsolePage.goto();
+    await systemConsolePage.toBeVisible();
+
+    // # Go to PLUGINS > Plugin Starter Template
+    await systemConsolePage.page.getByRole('link', {name: 'Plugin Starter Template'}).click();
+
+    // # Enable the plugin
+    await page.getByTestId('PluginSettings.PluginStates.com+mattermost+plugin-starter-template.Enabletrue').check();
```
hanzei, line 17: And here

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,149 @@
...
+          MM_CLUSTERSETTINGS_READONLYCONFIG: false
+          MM_EMAILSETTINGS_SMTPSERVER: inbucket
+          MM_EMAILSETTINGS_SMTPPORT: 10025
+          MM_PLUGINSETTINGS_ENABLEUPLOADS: true
+          MM_PLUGINSETTINGS_AUTOMATICPREPACKAGEDPLUGINS: false
+          MM_SERVICESETTINGS_SITEURL: http://localhost:8065
+          MM_SQLSETTINGS_DATASOURCE: "postgres://mmuser:mostest@postgres:5432/mattermost_test?sslmode=disable&connect_timeout=10"
+          MM_SQLSETTINGS_DRIVERNAME: postgres
+
+        ports:
```
hanzei, line 59:
```suggestion
          MM_SQLSETTINGS_DRIVERNAME: postgres
        ports:
```

e2e/README.md
```diff
@@ -0,0 +1,41 @@
...
+-   Build and deploy plugin - `make deploy`
+
+---
+
+### Run the tests
+
+Start Mattermost server:
+
+-   `cd <path>/mattermost/server`
+-   `make test-data`
```
hanzei, line 21: Do CI also need to run that command?

e2e/setup-environment.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+echo "Clone mattermost"
+cd ../../
+git clone --depth 1 https://github.com/mattermost/mattermost.git
```
hanzei, line 4: What happens if a plugin dev runs the script a second time? Does git pull the latest changes?

hanzei (changes requested):
Great work @saturninoabril :rocket: 

Looking forward to more plugins with E2E tests!

.github/workflows/e2e.yml
saturninoabril: Sure, I'll use the `actions/checkout` for consistency with the default depth and branch.

.github/workflows/e2e.yml
saturninoabril: I suggest to do this manually. I'll add notes in addition to what files to edit in the README.

e2e/.nvmrc
saturninoabril: It would be best to upgrade this matching the one being used at monorepo's webapp. However, I've seen issues in upgrading webapp's node version here and so it would be good to address separately.

Note that the Playwright from the monorepo uses a feature (native fetch) available from 18+ and so, use of outdated node 16 will not work.

I'll take a look at the options here.

e2e/.prettierrc.json
saturninoabril: Yup, sounds good.

e2e/support/constant.ts
saturninoabril: Same -- to do it manually with guides at the readme. This is base test only and will fail if not change.

e2e/tests/basic.spec.ts
saturninoabril: Same as above.

e2e/setup-environment.sh
saturninoabril: It will fail in cloning the repo (remain as is and will not pull the latest) with the following log but will continue until the end.
```
fatal: destination path 'mattermost' already exists and is not an empty directory.
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,148 @@
...
+        run: |
+          cd e2e
+          mkdir results
+          mv playwright-report results
+
+      - uses: actions/upload-artifact@26f96dfa697d77e81fd5907df203aa23a56210a8 # v4.3.0
+        if: success() || failure()
+        with:
+          name: test-results
+          path: e2e/results
```
saturninoabril, line 148: I'll move this as a reusable workflow at https://github.com/mattermost/actions-workflows. With that, I'm putting reviews on hold for now.

e2e/README.md
saturninoabril: Not needed since Playwright generate test data depending on test requirement (on a per spec basis)

e2e/setup-environment.sh
hanzei: It would be nice if the script runs without an error and pulls the latest changes on a second run.

saturninoabril: I agree, but some may prefer to avoid introducing unintended changes to the work they're currently doing in the monorepo. This script is intended solely for the initial setup to help users get started quickly. For subsequent steps, they can choose how they want to update their setup once they're more familiar with the dependencies.

pavelzeman:
GIthub clean up - this has been opened for a while, are we still doing this?
Would it make sense to either close or move to Draft?

saturninoabril:
Thanks Pavel, and closing this PR for now.

PR #211, opened 2024-10-17
Add emotion/react dependency

grubbins:
#### Summary

Adds `emotion/react` to the webapp dependencies.

When migrating a plugin to the latest template, I found it had a missing dependency on `emotion/react`, referencing this line:
```javascript
export default class UserAttribute extends React.PureComponent {
```

Seems worth adding to the `package.json`. I picked a version of similar age to `emotion/core` and all seemed well.


wiggin77 (approved):

wiggin77:
Thanks @grubbins

PR #212, opened 2024-11-08
Add boilerplate for api, commands, kvstore access and job management

BenCookie95:
#### Summary
**Problem:**
The starter template is too bare bones to kickstart most plugin efforts, requiring a lot of copy and paste from “known good” plugins to achieve a stable starting point.

**Justification:**
DE doesn’t use the starter template all that often anymore, spending extra time bootstrapping a plugin with all the necessary plumbing. This also leads to unnecessary divergences between plugins, complicating our developer integration strategy and story.

I added some skeleton code for creating an api and adding new slash commands. I separated the slash commands and KVStore access into their own packages for ease of use. I also added a simple mock for the command package to show how these separate packages can be easily tested. 

I noticed we used this Job Manager code across a few different plugins and I thought it would be very useful to bring into the skeleton.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61468
https://mattermost.atlassian.net/browse/MM-60911
https://mattermost.atlassian.net/browse/MM-60910
https://mattermost.atlassian.net/browse/MM-60909


hanzei (commented):
I really appreciate the work you are doing for the starter template. This is a great way for community members to build plugins more easily.

I wonder if we should make the boilerplate code opt-in. Something like a cli tool that generates a fresh repo with all the code that was specified. go templates could drive such a CLI tool.

BenCookie95:
> I really appreciate the work you are doing for the starter template. This is a great way for community members to build plugins more easily.
> 
> I wonder if we should make the boilerplate code opt-in. Something like a cli tool that generates a fresh repo with all the code that was specified. go templates could drive such a CLI tool.

Go templates is a great suggestions and maybe something we can use in future! I know that too much boilerplate can be annoying but I think this current implementation is easy to manage because most stuff is separated into separate packages and it's just a case of deleting the package and the initialization logic in plugin.go.

server/api.go
```diff
@@ -0,0 +1,40 @@
...
+	router.Use(p.MattermostAuthorizationRequired)
+
+	apiRouter := router.PathPrefix("/api/v1").Subrouter()
+
+	apiRouter.HandleFunc("/hello", p.HelloWorld).Methods(http.MethodGet)
+
+	router.ServeHTTP(w, r)
+}
+
+func (p *Plugin) MattermostAuthorizationRequired(next http.Handler) http.Handler {
```
lieut-data, line 24: It would be awesome to hoist this into the `public` package.

server/api.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+func (p *Plugin) HelloWorld(w http.ResponseWriter, r *http.Request) {
+	if _, err := w.Write([]byte("Hello, world!")); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
```
lieut-data, line 39: I'd love to see logging illustrated here (odds are good the `http.Error` will fail if `w.Write` actually failed at this point).

server/command/command.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/pluginapi"
+)
+
+type Handler struct {
+	client *pluginapi.Client
+}
+
+type Command interface {
+	Handle(args *model.CommandArgs) (*model.CommandResponse, *model.AppError)
```
lieut-data, line 16: Can we expunge `model.AppError` in favour of `error`? I realize that requires a shim in the actual hook, but it would be ideal not to propagate this anywhere we don't strictly need it.

server/command/command.go
```diff
@@ -0,0 +1,63 @@
...
+	executeHelloCommand(args *model.CommandArgs) *model.CommandResponse
+}
+
+const helloCommandTrigger = "hello"
+
+// Register all your slash commands in the NewCommandHandler function.
+func NewCommandHandler(client *pluginapi.Client) Command {
```
lieut-data, line 23: One day, it would be awesome to hoist this into the `public` package and just register callbacks for the slash commands to be handled.

In that future, I wonder if we would need a discrete package at all and could just do this all in `command.go`?

server/command/command.go
```diff
@@ -0,0 +1,63 @@
...
+	err := client.SlashCommand.Register(&model.Command{
+		Trigger:          helloCommandTrigger,
+		AutoComplete:     true,
+		AutoCompleteDesc: "Say hello to someone",
+		AutoCompleteHint: "<username>",
```
lieut-data, line 28: I wonder if we should illustrate the advanced autocomplete functionality?

server/config/configuration.go
```diff
@@ -0,0 +1,22 @@
+package config
```
lieut-data, line 1: I'm not a big fan of pushing this into a discrete package, especially when it's just one file right now.

server/jobs/job_manager.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/mattermost/mattermost-plugin-starter-template/server/config"
+)
+
+type Job interface {
+	GetID() string
+	OnConfigurationChange(cfg *config.Configuration) error
+	Stop(timeout time.Duration) error
+}
+
+type JobManager struct {
```
lieut-data, line 19: Is this basically a goroutine manager for background work? I started reviewing thinking this was based on https://github.com/mattermost/mattermost/blob/master/server/public/pluginapi/cluster/job.go, but then I actually don't know when I would use this. Can you elaborate on the use case for a non-HA aware job manager?

server/jobs/startertemplate_job.go
```diff
@@ -0,0 +1,169 @@
...
+		kvstore: kvstore,
+	}, nil
+}
+
+func (j *StarterTemplateJob) GetID() string {
+	return j.id
+}
+
+// OnConfigurationChange is called by the job manager whenenver the plugin settings have changed.
+// It is suggested that you stop the current job (if any) and start a new job (if enabled) with new settings.
```
lieut-data, line 52: Architecturally: why? If it's a long running job, odds are decent it's halfway done and I'm not sure of the semantics of interrupting it midway (especially if it persists back to the database or filestore). Should we just let it finish?

lieut-data (changes requested):
Thanks, @BenCookie95! A few of my inline comments are more exploratory and aren't blocking, but a few high level thoughts I'd love to work through:
* I'm not a big fan of packages for packages sake, e.g. `configuration`. Similarly, `command`, and especially the resulting interface just makes navigating the code all that much harder. Could we hoist some of the common code into the `public` package and make this simple enough to have all the code live in the `main` package?
* The job manager feels like it comes with a lot of baggage that I'd love to avoid adding to the template. Could we illustrate running an HA-aware job using the existing `cluster` package instead?

lieut-data:
> I know that too much boilerplate can be annoying but I think this current implementation is easy to manage because most stuff is separated into separate packages and it's just a case of deleting the package and the initialization logic in plugin.go.

:+1: to just having code vs. generating. That being said, as noted above, I'm not a huge fan of packages achieve this, so hopefully we can find a middle ground where it's still easy to remove but doesn't come with lots of cognitive overhead in the way the plugin is structured.

server/api.go
BenCookie95: Yeah that makes sense

server/api.go
BenCookie95: Agreed, missed the chance to log here

server/command/command.go
BenCookie95: Yeah good catch

server/plugin.go
```diff
@@ -1,28 +1,140 @@
...
 
 	// configuration is the active plugin configuration. Consult getConfiguration and
 	// setConfiguration for usage.
-	configuration *configuration
+	configuration *config.Configuration
+}
+
+func (p *Plugin) OnInstall() error {
+	return nil
+}
```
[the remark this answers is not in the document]
BenCookie95: You're right, I was confused by what it meant by "onboarding", I thought it was more general purpose. I'll remove it from the boilerplate

BenCookie95:
> I'm not a big fan of packages for packages sake, e.g. configuration. Similarly, command, and especially the resulting interface just makes navigating the code all that much harder. Could we hoist some of the common code into the public package and make this simple enough to have all the code live in the main package?

My main thought process of creating separate packages from the start was to discourage the overuse of the `main` package. I've read through a bunch of plugins and it gets very hard to navigate or figure out where anything is when it lives in the main package. You mentioned that you aren't a fan of using packages due to the cognitive overhead in the way the plugin is structured. Am I alone in wanting more packages to achieve this? I personally find it much easier to read and navigate.

 I agree that I could move the config back into the main package because it doesn't serve much use as it's own package (probably went overboard with that one). Command in it's current state made sense to be in a separate package because there is a fair amount of boilerplate to work through. I'll take a look at hoisting some of that to the public package and then it would probably make sense to move it back to main?

> The job manager feels like it comes with a lot of baggage that I'd love to avoid adding to the template. Could we illustrate running an HA-aware job using the existing cluster package instead?

Yeah, my initial thoughts were from a DE perspective because this was reused in a few plugins but it's overkill for a contributor coming in. I'll simplify it

server/command/command.go
BenCookie95: What is that advanced functionality? I'm not aware, is there something I can read?

BenCookie95:
Resolved most of the comments. Removed the config package and cleared out the overly complicated Job stuff

lieut-data:
> My main thought process of creating separate packages from the start was to discourage the overuse of the main package. I've read through a bunch of plugins and it gets very hard to navigate or figure out where anything is when it lives in the main package. You mentioned that you aren't a fan of using packages due to the cognitive overhead in the way the plugin is structured. Am I alone in wanting more packages to achieve this? I personally find it much easier to read and navigate.

@BenCookie95, for context, I recently went through an exercise in MS Teams collapsing many of the packages back into the `main` package. Most of the secondary packages were tightly coupled with the `Plugin` and required lots of interface juggling when more complex business logic was required. I was able to eliminate maybe 25% of the code overhead and make the code easier to navigate without unhelpful interfaces. I'm curious how you might find reading https://github.com/mattermost/mattermost-plugin-msteams/tree/main/server (not that it's perfect by any stretch!)

On the "when" side for new packages, I was chatting with @crspeller and wrote down a few shared thoughts:
* **Fewer packages is better**: default to the `main` package unless there's good reason for a new package.
* **Coupling implies same package**: don't jump through hoops to break apart code that's naturally coupled.
* **New package for a new interface**: a classic example is the `sqlstore` with layers for monitoring performance, caching and mocking.
* **New package for upstream integration**: a discrete `client` package for interfacing with a 3rd party is often a great place to break out into a new package

Curious as to your thoughts?

server/command/command.go
lieut-data: Take a peek at https://github.com/mattermost/mattermost-plugin-playbooks/blob/6c50600fb6151ea1abfed82961e980131b0ca2ab/server/command/command.go#L61 (ironically in an unnecessary `command` package itself ;P)

server/configuration.go
```diff
@@ -30,16 +30,16 @@ func (c *configuration) Clone() *configuration {
...
-		return &configuration{}
-	}
+// 	if p.configuration == nil {
+// 		return &configuration{}
+// 	}
 
-	return p.configuration
-}
+// 	return p.configuration
+// }
```
lieut-data, line 42: Intentional?

server/plugin.go
```diff
@@ -20,9 +37,46 @@ type Plugin struct {
...
-	fmt.Fprint(w, "Hello, world!")
+// OnActivate is invoked when the plugin is activated. If an error is returned, the plugin will be deactivated.
+func (p *Plugin) OnActivate() error {
+	p.client = pluginapi.NewClient(p.API, p.Driver)
+
+	p.kvstore = kvstore.NewKVStore(p.client)
+
+	p.commandClient = command.NewCommandHandler(p.client)
+
+	job, cronErr := cluster.Schedule(
```
lieut-data, line 48: nit: do we need `cronErr` or can we encourage universal adoption of just `err`?

server/job.go
```diff
@@ -0,0 +1,5 @@
+package main
+
+func (p *Plugin) runJob() {
+	// Include job logic here
```
lieut-data, line 4: Maybe log something to show the job running one an hour?

server/api.go
```diff
@@ -0,0 +1,41 @@
+package main
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost/server/public/plugin"
+)
+
+// ServeHTTP demonstrates a plugin that handles HTTP requests by greeting the world.
```
lieut-data, line 10: Propose we document the fully qualified path to hit this endpoint -- this came up in a recent hackathon with a prospect who asked how to hit the API.

lieut-data (approved):
Thanks @BenCookie95! One comment about the commented code, but otherwise non-blocking.

BenCookie95:
> > My main thought process of creating separate packages from the start was to discourage the overuse of the main package. I've read through a bunch of plugins and it gets very hard to navigate or figure out where anything is when it lives in the main package. You mentioned that you aren't a fan of using packages due to the cognitive overhead in the way the plugin is structured. Am I alone in wanting more packages to achieve this? I personally find it much easier to read and navigate.
> 
> @BenCookie95, for context, I recently went through an exercise in MS Teams collapsing many of the packages back into the `main` package. Most of the secondary packages were tightly coupled with the `Plugin` and required lots of interface juggling when more complex business logic was required. I was able to eliminate maybe 25% of the code overhead and make the code easier to navigate without unhelpful interfaces. I'm curious how you might find reading https://github.com/mattermost/mattermost-plugin-msteams/tree/main/server (not that it's perfect by any stretch!)
> 
> On the "when" side for new packages, I was chatting with @crspeller and wrote down a few shared thoughts:
> 
> * **Fewer packages is better**: default to the `main` package unless there's good reason for a new package.
> * **Coupling implies same package**: don't jump through hoops to break apart code that's naturally coupled.
> * **New package for a new interface**: a classic example is the `sqlstore` with layers for monitoring performance, caching and mocking.
> * **New package for upstream integration**: a discrete `client` package for interfacing with a 3rd party is often a great place to break out into a new package
> 
> Curious as to your thoughts?

I'm happy with that criteria you laid out and I think I'll add it to the README of this plugin so contributors have some guidance. Thanks! 

> required lots of interface juggling when more complex business logic was required

I can see how that can be painful if overused

README.md
```diff
@@ -64,6 +64,26 @@ To avoid having to manually install your plugin, build and deploy your plugin us
...
+
+#### Command package
+
+This package contains the boilerplate for adding a slash command and an instance of it is created in the `OnActivate` hook in plugin.go. If you don't need it you can delete the package and remove any reference to `commandClient` in plugin.go. The package also contains an example of how to create a mock for testing.
+
+#### KVStore package
+
+This is a central place for you to access the KVStore methods that are available in the `pluginapi.Client`. The package contains an interface for you to define your methods that will wrap the KVStore methods. An instance of the KVStore is created in the `OnActivate` hook.
+
+#### Jobs package
```
wiggin77, line 83: Not a package anymore.

wiggin77 (approved):
LGTM :+1:   One change needed to readme.

PR #217, opened 2025-03-25
Update to go1.24

hanzei:


fmartingr (approved):
This looks good to me, but two questions tho:

Should we keep the plugins go version one major version behind the same way we do in the Mattermost server or are we deferring that decision to the plugin maintainers? 

Should we involve prodsec in the approval (as per the above)? (Not sure since this is not an actual plugin, but will became several plugins over time).

hanzei:
> Should we keep the plugins go version one major version behind the same way we do in the Mattermost server or are we deferring that decision to the plugin maintainers?

Due to the long development cycles, new plugins will (naturally) fall behind in Go versions. By the time a plugin is released, the Go release cycle has moved forward significantly. I'm more concerned about outdated Go versions then (too) new ones. Hence, the push to update early.

@enzowritescode Thoughts?

enzowritescode:
2/5 I prefer if the versions are the same everywhere so we aren't having to keep track of multiple versions of everything. It makes managing potential vulnerabilities easier. It may also prevent confusion for everyone working on multiple projects not having to worry about the Go version and if the thing they want to use is available or not in that version.

I think there is some historical knowledge that I am missing on the versioning selection though so maybe @esarafianou or @jupenur can chime in.

hanzei:
cc @streamer45 as worked on go version updates recently

streamer45:
I agree on being consistent. We could probably add a step to take care of this to our [existing Playbook](https://hub.mattermost.com/playbooks/playbooks/k6xrudry9fgkxk3mktffyhiega). /cc @agarciamontoro

The existing strategy is to always be one version behind the tip. So when 1.24 went out, we updated our pipelines to 1.23. The rationale is to limit breaking changes and potential security vulnerabilities that often come in new versions.

wiggin77:
+1 on alignment with MM server.  I think breaking changes in latest Go is more likely to affect us than being on slightly older versions.

agarciamontoro:
I agree that consistency is important, and having to change Go versions between projects we control would be a bit painful.

> We could probably add a step to take care of this to our [existing Playbook](https://hub.mattermost.com/playbooks/playbooks/k6xrudry9fgkxk3mktffyhiega). /cc @agarciamontoro

@streamer45, do you mean updating plugins in general? This starter template in particular?

esarafianou:
+1 on alignment with MM server. Claudio described the existing strategy and the rationale perfectly.

lieut-data:
No objection to matching the MM server, but one note:

> I agree that consistency is important, and having to change Go versions between projects we control would be a bit painful.

Is this an issue in practice, @agarciamontoro, after https://go.dev/doc/toolchain?

wiggin77:
> By putting the Starter Template one version ahead of the MM server, we give Plugins a head start - which they need.

I don't think this is really an issue.  It is not hard to upgrade the Go version in any plugin.  I think adopting the latest Go version too early will bite us far more often than upgrading slowly, despite one potential example outlier.  I like the one version behind policy we have now, and I think the starter-template should sync with server.

hanzei:
Ci seems to be broken by https://github.com/mattermost/actions-workflows/pull/71

hanzei:
The server [has been updated to go1.24](https://github.com/mattermost/mattermost/pull/31220). I think we are good to proceed with this PR.

lieut-data (changes requested):
Actually, I just tried doing this on the monorepo and realized it merges the dependencies of the tools into the core `go.mod`. I'm worried this increases the surface area we have to track for security vulnerabilities, plus we're doing version management on a larger set of binaries, some of which are outside of our control.

It seems like the Go team identified this concern, but didn't address it with v1.24. @agnivade, thoughts on this? I think we should be consistent one way or another: if it's ok, let's do it everywhere.

agnivade:
Ahh sorry, I didn't even read the PR message properly. I missed that this PR was about the `go tool` command _and_ upgrading Go toolchain. 

I don't follow Go design discussions that closely now. But this seems like a more formalized implementation of the go.tools.mod file we had in the past. We moved from that to simple `go install binary@version` commands because they were simpler to maintain.

There might be cases where a company wants to track dependencies of tools more closely, but I think the cost we are paying by having to track all tool dependencies in our go.mod is higher than the benefit of reducing a few lines in Makefile.

I'd be of the opinion to keep using `go install` to install our tools, and keep our go.mod clean.

lieut-data:
I'm 👍🏼 for the `tools.mod` file, seems elegant. 

One slightly weird thing about this setup is that I don't understand why the `tool` block doesn't specify versions like the require block does:

```go
tool (
	github.com/golang/mock/mockgen
	github.com/golangci/golangci-lint/cmd/golangci-lint
	gotest.tools/gotestsum
)

require (
	github.com/golang/mock v1.6.0
	github.com/gorilla/mux v1.8.1
	github.com/mattermost/mattermost/server/public v0.1.15
	github.com/pkg/errors v0.9.1
	github.com/stretchr/testify v1.10.0
)
```

Instead, I have to dig into the indirect dependencies to find a version:
```
github.com/golangci/golangci-lint v1.64.8 // indirect
```

I don't think this is a blocker - just weird :/

agnivade:
The tools.mod approach looks good to me. Thanks Ben :+1:

lieut-data:
Regrettably, it looks like this approach won't work on the monorepo for now due to our use of workspaces: https://github.com/golang/go/issues/73968

agnivade:
Ahhh :sob:

hanzei:
I'm going to close this PR. It doesn't seem like Tool Dependencies are ready for wide use within Mattermost.

hanzei:
Here is a PR to update to go1.24 without the tooling changes: https://github.com/mattermost/mattermost-plugin-starter-template/pull/225

PR #221, opened 2025-05-15
Upgrade `@mattermost/client`, `@mattermost/types` and `mattermost-redux` to v10.7.0

danilvalov:
Fixed PR for https://github.com/mattermost/mattermost-plugin-starter-template/pull/216


webapp/webpack.config.js
```diff
@@ -36,6 +36,9 @@ const config = {
     resolve: {
         alias: {
             '@': path.resolve(__dirname, 'src'),
+            // temporary fix for `mattermost-redux` v10+
+            // see more: https://github.com/mattermost/mattermost/issues/31060
+            'mattermost-redux': 'node_modules/mattermost-redux/lib',
```
hmhealey, line 41: Good fix! I suppose it'll be a bit less temporary if anyone uses this template in the time before it's removed, but that should be fine since this is at worst just redundant

hmhealey (approved):

webapp/webpack.config.js
BenCookie95: We are going to fix https://github.com/mattermost/mattermost/issues/31060 internally so this won't be needed. I'm going to hold off approving this as I'd like to wait for the internal fix

BenCookie95 (changes requested):
Thanks @danilvalov for these updates, I appreciate the workaround you figured out. We are going to fix the issue at it's source though. I'll comment on this PR when that's merged

hmhealey (changes requested):
@danilvalov Could you upgrade these libraries to the 10.8.0 versions we just published and remove the workaround in the Webpack config? This new version should work without the workaround, and I don't think we'll need any other changes

esarafianou:
@hmhealey you can ignore the Snyk check

PR #226, opened 2025-07-31
Disable eslint header rule

grubbins:
The imported eslint config tells eslint that we should add a mattermost copyright header to every file. For users developing their own plugins, this should be disabled.


lieut-data (approved):

fmartingr (approved):
LGTM. Since the eslint header plugin is not explicitly specified in the `package.json`, it is extending the config in the `mattermost/eslint-plugin` config?

PR #234, opened 2025-11-11
MM-66286 Upgrade React, Redux, and MM packages for React 18

hmhealey:
#### Summary
When I started this a month ago, I ran into some issues which I needed to address with this in mattermost-redux, but with those done, it's pretty straightforward on the plugin side

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66286


hanzei (approved):
LGTM 👍

coltoneshaw (approved):
lgtm

grubbins:
Is this meant to be applied in plugins in sync with upgrading to Mattermost 11?

I think in Mattermost 10, `window.React` is still `17.0.2`. Since this is configured as a webpack external, does that mean that plugins built with React 18 will actually get 17 at runtime?

PR #237, opened 2026-01-20
Update types for Plugin User Settings

danilvalov:
Now there are incorrect types for user settings.

Copied from: [github.com/mattermost/mattermost/webapp/channels/src/types/plugins/user_settings.ts](https://github.com/mattermost/mattermost/blob/master/webapp/channels/src/types/plugins/user_settings.ts)


hmhealey (approved):
@danilvalov Thanks for the PR to fix those! I want to start exporting those types from the web app eventually so that we won't need to worry about those getting out of sync

PR #240, opened 2026-02-23
Updating nvm version and Makefile version flakiness fix

avasconcelos114:
#### Summary

This PR syncs the plugin starter template with a few changes that happened in downstream plugins that needed to be propagated here as well
* Bumping Node version to 24.13.1
* Added Makefile fix in `make major` / `make minor` / `make patch` commands that fixes some flakiness with iterating to a following version after an RC (e.g. Having a `v1.4.0-rc1` generated from `make minor-rc` causes the following `make minor` command resolve to `v1.5.0` instead of the intended `v1.4.0`)


fmartingr (approved):

hanzei (approved):
Thanks!

PR #241, opened 2026-03-01
Add testcontainers-based integration test infrastructure

wiggin77:
#### Summary
Plugin authors currently have two options for testing server-side behavior: mock everything via plugintest.API or write E2E/Playwright tests. Mocks are fragile and don't test real behavior, they verify the plugin calls the right methods with the right arguments, but not that the plugin actually works against a real Mattermost server. E2E tests are slow, hard to write, and test browser-to-server rather than plugin-to-server.

This package fills the gap by giving every plugin a way to write Go integration tests that exercise the plugin through the real Mattermost API: KVStore, posts, channels, users, hooks, slash commands.  These get tested against an actual server with a real database. Since it lives in the starter template, every new plugin gets this infrastructure for free.

- Adds a server/testhelper/ package that spins up real Postgres + Mattermost containers via testcontainers, deploys the plugin, and provides an authenticated model.Client4 for testing against a live server                                                                                                                                                            

- Each Setup() call resets the database, restarts the Mattermost container (re-initializing default roles/permissions), re-deploys the plugin, and creates a fresh team/user/channel, guaranteeing full isolation between tests                                                                                                                                                

- make test now depends on make dist with BUILD_FOR_TEST=1, building only host-arch instead of all 5 platform targets                                                            

- Includes unit tests for the testhelper's pure functions and integration tests validating the infrastructure itself (container startup, DB reset isolation, user/channel/post creation, plugin deployment)

#### Ticket Link
NONE


fmartingr:
Should we be opinionated and put this into a package importable by all plugins? Or if we are going the route of updating the plugins with AI so having duplicates it's not a problem anymore, maybe putting this outside the server package may help. We still need to remember to commit changes we made to things like this to upstream.

wiggin77:
> Should we be opinionated and put this into a package importable by all plugins? Or if we are going the route of updating the plugins with AI so having duplicates it's not a problem anymore, maybe putting this outside the server package may help. We still need to remember to commit changes we made to things like this to upstream.

Thanks @fmartingr I put it here for discoverability and based on the knowledge that we will have a way to keep plugins updated with starter-template changes.  Technically it is importable where it is, although it would be confusing to import a package you already have.  

If we want to put this somewhere it can be imported, then we should make the whole starter template just a stub that imports all the rest.

fmartingr:
> > Should we be opinionated and put this into a package importable by all plugins? Or if we are going the route of updating the plugins with AI so having duplicates it's not a problem anymore, maybe putting this outside the server package may help. We still need to remember to commit changes we made to things like this to upstream.
> 
> Thanks @fmartingr I put it here for discoverability and based on the knowledge that we will have a way to keep plugins updated with starter-template changes. Technically it is importable where it is, although it would be confusing to import a package you already have.
> 
> If we want to put this somewhere it can be imported, then we should make the whole starter template just a stub that imports all the rest.

I don't mind the code in the starter-template, just split between the `server` package (scaffolding) and another one that holds shared tooling that either do not belong to the public package or one that we can iterate faster on. Not a blocker tho, just a suggestion.

wiggin77:
> just split between the `server` package (scaffolding) and another one that holds shared tooling

@fmartingr  I like that this test tooling is completely contained in a package and not sprinkled around.  It's in the server package because it only applies to server unit tests.  I'm open to putting it somewhere else, but if we do, we should bring all the plugin tooling from the starter-template with it so we can update easily.

Makefile
```diff
@@ -218,11 +220,17 @@ ifneq ($(MM_SERVICESETTINGS_ENABLEDEVELOPER),)
 	@echo Building plugin only for $(DEFAULT_GOOS)-$(DEFAULT_GOARCH) because MM_SERVICESETTINGS_ENABLEDEVELOPER is enabled
 	cd server && env CGO_ENABLED=0 $(GO) build $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-$(DEFAULT_GOOS)-$(DEFAULT_GOARCH);
 else
-	cd server && env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-linux-amd64;
-	cd server && env CGO_ENABLED=0 GOOS=linux GOARCH=arm64 $(GO) build $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-linux-arm64;
-	cd server && env CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 $(GO) build $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-darwin-amd64;
-	cd server && env CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 $(GO) build $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-darwin-arm64;
-	cd server && env CGO_ENABLED=0 GOOS=windows GOARCH=amd64 $(GO) build $(GO_BUILD_FLAGS) $(GO_BUILD_GCFLAGS) -trimpath -o dist/plugin-windows-amd64.exe;
+	@if [ "$(BUILD_FOR_TEST)" = "1" ]; then \
```
fmartingr, line 223: What the benefit of this against `MM_SERVICESETTINGS_ENABLEDEVELOPER`?

Makefile
```diff
@@ -218,11 +220,17 @@ ifneq ($(MM_SERVICESETTINGS_ENABLEDEVELOPER),)
...
+		echo "Building plugin only for linux-$(DEFAULT_GOARCH) for integration tests"; \
```
fmartingr, line 224: Why only for linux?

hanzei (commented):
> we should bring all the plugin tooling from the starter-template with it so we can update easily.

That is something I would love to do moving forward. But some non-trival challenges with the various tools @fmartingr and I tested.

This new library can fully live in a shared place without interfering with existing plugins. We did experience pain in the past by not sharing common tools among plugins. I suggest we learn from that and do it right from the start this time.

wiggin77:
> > we should bring all the plugin tooling from the starter-template with it so we can update easily.
> 
> That is something I would love to do moving forward. But some non-trival challenges with the various tools @fmartingr and I tested.
> 
> This new library can fully live in a shared place without interfering with existing plugins. We did experience pain in the past by not sharing common tools among plugins. I suggest we learn from that and do it right from the start this time.

There is a lot of boiler-plate code that can be put in a separate repo and imported into plugins without any special tools.  Alternatively we can leave it in the starter template and provide a way to update starter template code, which is miles more complicated.  

I have a need for this right now and have imported this PR commit into my plugins, which is not ideal.  @hanzei @fmartingr if you have a concrete proposal we can implement it, otherwise we can leave it as is in the PR and change it later.

fmartingr:
> > > we should bring all the plugin tooling from the starter-template with it so we can update easily.
> > 
> > 
> > That is something I would love to do moving forward. But some non-trival challenges with the various tools @fmartingr and I tested.
> > This new library can fully live in a shared place without interfering with existing plugins. We did experience pain in the past by not sharing common tools among plugins. I suggest we learn from that and do it right from the start this time.
> 
> There is a lot of boiler-plate code that can be put in a separate repo and imported into plugins without any special tools. Alternatively we can leave it in the starter template and provide a way to update starter template code, which is miles more complicated.
> 
> I have a need for this right now and have imported this PR commit into my plugins, which is not ideal. @hanzei @fmartingr if you have a concrete proposal we can implement it, otherwise we can leave it as is in the PR and change it later.

Once this is merged, would you be happy to import the starter-template into other plugins? I don't mind having to import the starter template to get toolings (we could to the same for the manifest and pluginctl binaries, among other things). This could also live under a `pluginsdk` package in the monorepo, but that have other complexities. Having this here allows us to iterate faster but we need to start versioning the starter-template more often then.

wiggin77:
> would you be happy to import the starter-template into other plugins?

@fmartingr it is not ideal but until a centralized plugin SDK with upgrade capabilities is created this will do.

hanzei:
@wiggin77 I've created https://github.com/mattermost/mattermost/pull/35643. You can start using it right away.  Once the PR is merged, I can create a new release of `public`.

fmartingr:
Removed myself from reviewers while the discussion of the location of this feature is ongoing. Happy to test and review it on its final place.

docs/TESTCONTAINER_PLAN.md
hanzei, line 1: We want to remove this before merge, right?

Makefile
```diff
@@ -341,8 +349,11 @@ detach: setup-attach
 	fi
 
 ## Runs any lints and unit tests defined for the server and webapp, if they exist.
+## Depends on dist so the plugin bundle is available for integration tests.
+## BUILD_FOR_TEST=1 restricts the server build to linux/$(DEFAULT_GOARCH) only.
 .PHONY: test
-test: apply webapp/node_modules install-go-tools
+test: BUILD_FOR_TEST = 1
+test: dist install-go-tools
```
hanzei, line 356: Why where `apply webapp/node_modules` removed?

Makefile
```diff
@@ -169,7 +169,7 @@ major-rc: ## to bump major release candidate version (semver)
 
 ## Checks the code style, tests, builds and bundles the plugin.
 .PHONY: all
-all: check-style test dist
+all: check-style test
```
hanzei, line 172: Unrelated change?

hanzei (changes requested):
While I would still appreciate if the test helper would live in the public module, it's not a blocker for this PR.

A few questions below

hanzei:
And sorry for the delay @wiggin77 :pray:

docs/TESTCONTAINER_PLAN.md
wiggin77: Some projects are keeping plans and others don't.  I've moved it to /docs/.

Makefile
wiggin77: `webapp/node_modules` is applied via `dist` -> `webapp`.

Makefile
wiggin77: Now that test depends on `dist`, having `dist` here too would just build it a second time.

wiggin77:
@fmartingr please weigh in on where this library should live.

wiggin77:
@coderabbitai review

go.mod
```diff
@@ -1,33 +1,58 @@
...
-go 1.25
+go 1.25.0
 
 require (
 	github.com/gorilla/mux v1.8.1
 	github.com/mattermost/mattermost/server/public v0.1.21
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.11.1
+	github.com/testcontainers/testcontainers-go v0.40.0
+	github.com/testcontainers/testcontainers-go/modules/postgres v0.40.0
```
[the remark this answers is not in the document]
wiggin77: `docker/docker v28.5.1+incompatible` is a test-only indirect dependency pulled by `testcontainers-go`, which is already at the latest release (v0.40.0). The `+incompatible` suffix only means the module predates semantic-import-versioning; it is not a vulnerability marker, and the finding cites no concrete CVE. It is not in the shipped plugin binary. No actionable, verified issue.

server/integration_test.go
```diff
@@ -0,0 +1,70 @@
...
+// then confirms the plugin reaches the Running state by querying the plugin status API.
+//
+// This is the most basic integration test — if this fails, the plugin cannot activate at all
+// (e.g., OnActivate returns an error, missing dependencies, manifest issues).
+func TestPluginActivation(t *testing.T) {
+	th := testhelper.Setup(t)
+
+	ctx := context.Background()
+	statuses, _, err := th.AdminClient.GetPluginStatuses(ctx)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
wiggin77: `go test` already bounds hangs via its global -timeout, and Setup() uses container wait strategies with explicit startup timeouts. Adding per-call deadlines is optional hardening, not a correctness bug, and the template stays simpler without it.

server/testhelper/mmcontainer.go
```diff
@@ -0,0 +1,232 @@
...
+
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/postgres"
+	"github.com/testcontainers/testcontainers-go/network"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+const (
+	defaultMMImage = "mattermost/mattermost-enterprise-edition:latest"
+
```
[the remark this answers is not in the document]
wiggin77: Using `latest` is a deliberate maintenance choice for a starter template (catch breakage against current MM).

server/testhelper/mmcontainer.go
```diff
@@ -0,0 +1,232 @@
...
+	mmContainer   testcontainers.Container
+	dockerNetwork *testcontainers.DockerNetwork
+	serverURL     string
+}
+
+var (
+	sharedContainers *mmContainers
+	containersOnce   sync.Once
+	containersErr    error
+)
```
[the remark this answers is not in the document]
wiggin77: A mutex would not make parallel tests correct (one test's reset wipes another's data mid-run); the design is inherently serial. The right minimal fix is to document that Setup() must not be used with t.Parallel().

server/testhelper/mmcontainer.go
```diff
@@ -0,0 +1,232 @@
...
+			return
+		}
+
+		sharedContainers = &mmContainers{
+			pgContainer:   pgC,
+			mmContainer:   mmC,
+			dockerNetwork: nw,
+			serverURL:     serverURL,
+		}
+	})
```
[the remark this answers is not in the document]
wiggin77: Wait strategies bound readiness and go test -timeout bounds the rest. Threading a bounded context through every Docker call is invasive for marginal benefit.

server/testhelper/mmcontainer.go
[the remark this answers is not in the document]
wiggin77: This is fixed

fmartingr:
> @fmartingr please weigh in on where this library should live.

I'd say it make sense to move it where the other plugin sdks resides, `public/plugine2e` ? Unsure where the conversation is at right now.

fmartingr (approved):

hanzei (commented):
Please check out https://github.com/mattermost/mattermost/pull/35643. I've moved the lib into the shared `plugin` package with https://github.com/mattermost/mattermost-plugin-starter-template/pull/249 exercising that new package.

A centralized package gives each plugin enough guardrails to implement new e2e tests without writing too much boilerplate first.

PR #243, opened 2026-04-03
Remove enzyme

grubbins:
The enzyme dependencies, and the related jest setup for snapshot support in tests, should be removed because:
1. They break the tests (see below) and
2. This pattern is quite out of date, and enzyme is barely supported on React 18. (Note that `enzyme-adapter-react-17-updated` was already removed under MM-66286, and there's no good React 18 version)

To reproduce the test failure from the current master branch, simply delete `package-lock.json` and then `make test`. This results in:
```
cd webapp && /opt/homebrew/bin/npm run test;

> test
> jest --forceExit --detectOpenHandles --verbose

 FAIL  src/react_fragment.test.tsx
  ● Test suite failed to run

    ReferenceError: ReadableStream is not defined

      at Object.<anonymous> (node_modules/undici/lib/web/webidl/index.js:534:58)
      at Object.<anonymous> (node_modules/undici/lib/web/fetch/util.js:12:20)

 FAIL  src/manifest.test.tsx
  ● Test suite failed to run

    ReferenceError: ReadableStream is not defined

      at Object.<anonymous> (node_modules/undici/lib/web/webidl/index.js:534:58)
      at Object.<anonymous> (node_modules/undici/lib/web/fetch/util.js:12:20)

Test Suites: 2 failed, 2 total
```

These errors arise via the `cheerio` library and then the `undici` library, a transitive dependency from the enzyme support. Unpicking these issues requires either some careful version pinning or some polyfilling. Alternatively, removing the enzyme dependencies as in this PR recovers working tests.

Since the enzyme snapshot pattern is very unlikely to be used in real plugin tests, I think it's better to remove enzyme.

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "webapp",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
@@ -27,7 +27,6 @@
         "@testing-library/jest-dom": "5.16.1",
         "@types/babel__core": "7.20.5",
         "@types/babel__template": "7.4.4",
-        "@types/enzyme": "3.10.11",
         "@types/jest": "27.4.0",
         "@types/node": "20.17.6",
         "@types/react": "18.2.64",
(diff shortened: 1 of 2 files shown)


grubbins:
I think the failed CI chain would be fixed by #244!

hanzei:
CI is green :tada:

BenCookie95 (approved):

PR #245, opened 2026-04-07
More mattermost-webapp types

grubbins:
Some more small improvements to the mattermost-webapp types file that needed to be made to resolve typescript errors in my plugins - plus a useful addition for plugins that use registerAdminConsoleCustomSetting.

I wonder if this file could be published as part of a library?

### DIFF
--- a/webapp/src/types/mattermost-webapp/index.d.ts
+++ b/webapp/src/types/mattermost-webapp/index.d.ts
@@ -3,10 +3,11 @@
 
 /* eslint-disable max-lines */
 
-import type {Reducer} from 'redux';
+import type {Reducer, UnknownAction} from 'redux';
 
 import type {WebSocketMessage} from '@mattermost/client';
 import type {Channel} from '@mattermost/types/channels';
+import type {AdminConfig, ClientLicense} from '@mattermost/types/config';
 import type {FileInfo} from '@mattermost/types/files';
 import type {Post, PostEmbed} from '@mattermost/types/posts';
 import type {ProductScope} from '@mattermost/types/products';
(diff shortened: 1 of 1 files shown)


webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -873,9 +894,9 @@ export interface PluginRegistry {
         }]
     ): {
         id: UniqueIdentifier;
-        showRHSPlugin: object;
-        hideRHSPlugin: object;
-        toggleRHSPlugin: object;
+        showRHSPlugin: UnknownAction;
```
grubbins, line 897: Strictly, this is inaccurate since these may be `ThunkAction`s instead, but I guess the difference is moot from the POV of the caller since the intent is just that the caller should `dispatch` them.

hanzei:
@coderabbitai review

webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -168,6 +168,25 @@ export type PluginConfigurationRadioSettingOption = {
...
+    config: Partial<AdminConfig>;
+    license: ClientLicense;
+    setByEnv: boolean;
+    onChange: PluginCustomSettingsComponentChangeHandler<ValueType>;
+    registerSaveAction: (saveAction: () => Promise<{error?: {message?: string}}>) => void;
+    setSaveNeeded: () => void;
+    unRegisterSaveAction: (saveAction: () => Promise<{error?: {message?: string}}>) => void;
+    cancelSubmit: () => void;
+    showConfirm: boolean;
+};
```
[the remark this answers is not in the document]
grubbins: I have used this type in a plugin and the lack of these imports did not cause any error, but it makes sense to add them anyway! Have done that

hanzei:
@coderabbitai review

webapp/src/types/mattermost-webapp/index.d.ts
hmhealey: That's probably for the best. I've spent enough time lately fighting lately with issues in having code that uses `mattermost-redux` dispatch thunk actions because of a hack I implemented previously. Ideally, we might type this more specifically, but I think this is good enough to tell users that it's an action which needs to be dispatched

hmhealey (approved):
@grubbins Thanks for the PR and for continuing to help make this easier to work with!

I've been working lately on a package to help share components between the web app and plugins, and while we can't include that here for general use yet (there's some compatibility concerns), I look forward to adding these types to it in the future to ensure that these types won't diverge from the web app in the future.

webapp/src/types/mattermost-webapp/index.d.ts
grubbins: Yes - typescript vs thunk has been interesting to learn. I also typically put this in a top-level `src/types` file in the plugins:

```
/// <reference types="mattermost-redux/types/extend_redux" />
```

That could also handy in the template IMO.

